PR #6, opened 2020-02-07
Lesson 1 Exercises

gildedgardenia:


src/chapters/data-types/exercises.rst
```diff
@@ -0,0 +1,64 @@
+Exercises: Data Types
+=====================
+
+To get started, open up Visual Studio and create a new console program by clicking *New* and selecting *Console Application* under *.Net Core App*.
```
CarlyLanglois: Can you make this more generic and/or add a todo to point to the part in the setup instructions to create a new program? im worried that mac-specific instructions won't translate to the windows setup

src/chapters/data-types/exercises.rst
```diff
@@ -0,0 +1,64 @@
...
+
+For each part of the exercises, carefully note in code comments which part of the exercises you are working on. 
+When you are done with that particular section of the exercises, comment out your solution and start working on the next one!
+
```
CarlyLanglois: I think it'd be good practice to have the students create a data-types-exercises solution and each of these items as a separate project. what do you think? that way they can keep each of their answers to the exercises in a working state - and they'll get practice running different projects in one solution and dealing with the Main methods/ program.cs

CarlyLanglois: the other thing is, we should talk again about the csharp-web-dev-exercises. I get now reading this taht it makes sense to have the students create their own solution program instead of adding projects to a starter repo, but maybe there's a packaging option like in java? i'll look into it

src/chapters/data-types/exercises.rst
```diff
@@ -0,0 +1,64 @@
...
+         string name = Console.ReadLine(); 
+
+   #. Use concatenation to print the greeting:
+
+      .. sourcecode:: csharp
+
+         Console.WriteLine("Hello " + name);
+
+   #. Click on the arrow in the top left to run your program.
```
CarlyLanglois: again, id be less specific about the arrow's location until we're more clear on the similarities/differences on windows

CarlyLanglois (approved):
looks great with some comments items it might be easier to chat about in person. 
I'd also suggest removing the other pages here in data types and setup since those will probably get merged after this and i made a lot of changes to the pagination and naming. or just merge in the latest that i pushed up this morning for both branches

src/chapters/data-types/exercises.rst
gildedgardenia: Done! Thanks!

src/chapters/data-types/exercises.rst
gildedgardenia: Done! Thanks!

src/chapters/data-types/exercises.rst
gildedgardenia: Done! Thanks!

CarlyLanglois (approved):
good shape - can you please merge in master and the latest changes from the branch this is based off before merging

PR #7, opened 2020-02-08
Lesson 1 Studio

gildedgardenia:


src/chapters/data-types/studio.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Some questions to ask yourself:
+
+#. What data type should the radius be?
+#. What is the best way to get user input into a variable ``radius`` of
+   that type?
+
+.. admonition:: Note
+
+   Use the ``Math`` class in C# to get the value of pi and square the radius.
```
CarlyLanglois: maybe we could refer to this as System.Math since we talked about the System namespace before and we haven't asked them to pull in another class from it yet, other than Console.

src/chapters/data-types/studio.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. admonition:: Note
+
+   Use the ``Math`` class in C# to get the value of pi and square the radius. 
+   The `documentation <https://docs.microsoft.com/en-us/dotnet/api/system.math?view=netframework-4.8>`_ has guidance on how to use the ``PI`` field and the ``Pow`` method.
+
+Add a Little Validation
+-----------------------
+
+Extend your program further by using a `while or do-while loop <https://www.w3schools.com/cs/cs_while_loop.asp>`__, so that when the user enters a negative number they are re-prompted for a radius.
```
CarlyLanglois: i think i remember you said you pulled this from the bonus missions yeah? I agree the studio could use something else but since we haven't introduced loops in cs yet, we should probably keep this idea as a bonus, like the other bonus that says they need to read ahead for conditional syntax. 
what about asking the students to provide units? or adding another output line about the circle's diameter/circumference or something?

CarlyLanglois (approved):
same thing i mentioned on the exercises -this will probs get merged before the branch its based off so kit might be easier to remove the files other than this studio and/or merge in the latest from the base branch.

otherwise great - with a suggestion a/b the second task!

src/chapters/data-types/studio.rst
CarlyLanglois: or hey, actually im writing the wiki now and from the java notes, i noticed that we ask students to break out their solution into more than one class. can we ask them to do that here too? as a second task?

src/chapters/data-types/studio.rst
gildedgardenia: Done! Thanks!

src/chapters/data-types/studio.rst
```diff
@@ -0,0 +1,54 @@
...
+
+.. admonition:: Note
+
+   Use the ``System.Math`` class in C# to get the value of pi and square the radius. 
+   The `documentation <https://docs.microsoft.com/en-us/dotnet/api/system.math?view=netframework-4.8>`_ has guidance on how to use the ``PI`` field and the ``Pow`` method.
+
+More Calculations
+-----------------
+
+#. Using the same radius calculate the circumference (``2*pi*r``) and diameter of the circle (``2*r``).
```
CarlyLanglois: Using the same radius,

src/chapters/data-types/studio.rst
```diff
@@ -0,0 +1,54 @@
...
+Road Trip!
+----------
+
+#. Ask the user for the miles per gallon of their car. 
+#. If the radius that they entered is in miles, output how many gallons of gas they will use to go around this circle. 
+
+Bonus Missions
+--------------
+
+#. Put the necessary properties and methods into a separate class.
```
CarlyLanglois: since nothing is exactly necessary to move to a different class at this point, maybe we could prompt them differently with something like, "Think about how we could make this program more modular by breaking out some of the code into a separate class. For example, we could pull out the circle information into a ``Circle`` class and leave the user questions and console messages in ``Program``.  Take a look at <link-to-hello-methods-example> for a refresher on using another class file."

CarlyLanglois (approved):
good again with another optional note on the first bonus mission. and be sure to square away with master before merging

src/chapters/data-types/studio.rst
gildedgardenia: Done!

src/chapters/data-types/studio.rst
gildedgardenia: Done!

PR #12, opened 2020-02-17
lsn 2 studio

CarlyLanglois:


src/chapters/control-flow-and-collections/studio.rst
```diff
@@ -0,0 +1,110 @@
...
+      char[] charactersInString = myString.ToCharArray();
+
+Some Items to Ponder Before Starting
+------------------------------------
+
+#. There are multiple ways to approach this task, but one way involves the
+   following steps:
+
+   a. Loop through the string one character at a time,
+   b. Store and/or update the count for a given character using an appropriate
```
gildedgardenia, line 22: I would change "an" to "the" since you later state that one data structure will be the best option out of the 3.

src/chapters/control-flow-and-collections/studio.rst
```diff
@@ -0,0 +1,110 @@
...
+      data structure.
+   c. Loop through the data structure to print the results (one character and its
+      count per line).
+
+#. Which type of data structure (``List``, ``array``, or ``Dictionary``)
+   should you use to store character counts? Any can be made to work, but there
+   is a BEST choice.
```
gildedgardenia: I might rephrase to "Any of these options would work, but using one of these data structures will optimize your code's efficiency." You could really use any effect, but the phrase "there is a BEST choice" reads a little awkward to me.

gildedgardenia (commented):
I think this is a good studio! I think the bonus missions are really fun and this is a good challenge. However, I felt like the instructions hop around a bit without explanation. Is there starter code or should the students start from scratch? Should they submit a link to a Github repo when they are done? I think a paragraph that transitions from "Before you get started" to "Here is the end result" would also help this.

src/chapters/control-flow-and-collections/studio.rst
CarlyLanglois: yes, i do like your phrasing better here! Id prefer to keep the ambiguity in the statement above, though. I like the idea that they can experiment with any of the collection types.

I added your re-write - lmk how that reads to you, keeping "an" above

CarlyLanglois (commented):
just pushed up changes from your notes. I re-arranged some items but left off the github instructions since we don't do formal studio tracking

PR #13, opened 2020-02-19
Assignment1

CarlyLanglois:
use https://github.com/LaunchCodeEducation/TechJobsConsole/pull/9 to pull down the starter code. lmk if you have to do anything for the build before you are able to run it

and there's a merge request for the solution code on gitlab


src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+pretty or have lots of features, but it will give you a chance to work through
+some initial concepts and get feedback from LaunchCode staff.
+
+Your mentor on this project is Blake.
+
+.. figure:: figures/LC-Blake.png
+   :scale: 50%
+   :alt: LaunchCode mentor Blake.
+
+Your LaunchCOde mentor, Blake.
```
gildedgardenia: "LaunchCode"

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+TechJobs (Console Edition)
+---------------------------
+
+The app you must help design is a simple console (i.e. command-line) prototype
+of the new TechJobs app. It will allow LaunchCode staff to browse and search
+listings of open jobs by employer partners.
+
+The prototype process gives everybody a chance to work out some initial ideas
+without investing a ton of time into developing a finished product. Once
+everybody likes the prototype, the Tech Team will begin work toward a
```
gildedgardenia: What about instead of "work toward", "working on"?

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+Before diving in and starting to code, make sure you understand what the code
+you’ve been given does. Since you’re starting with a functioning---albeit
+unfinished---program, go ahead and run it to get an idea of how it works. 
+
+.. admonition:: Warning
+
+   The application will run until you force it to quit, re-prompting time
+   after time. To kill it, enter *ctrl+C* in the terminal or use the square 
+   *stop* icon in the vicinity of the triangle *run* icon. We’ll learn precisely 
+   how the program manages to work this way below.
```
gildedgardenia, line 86: One note about this. When I attempted to force quit, ``dotnet`` shut down completely, prompting a warning box. If you have seen this before, I would let students know this can happen. If not, probably just my machine being weird!

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+Let’s explore the code by starting with the source of the data our program is
+providing access to.
+
+The Data File: ``jobs_data.csv``
+--------------------------------
+
+Our simple app doesn’t connect to a database. If the prototype proves
+useful and we continue development, we’ll add that functionality later.
+But for now, we’ve been given a CSV (comma-separated values) file from
+the Company Team at LaunchCode that contains some recent jobs. This file
```
gildedgardenia: Maybe "jobs" to "job openings"

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+was exported from an Excel spreadsheet into this format, which is easy
+for programs to read in.
+
+If CSV files are new to you, don’t worry, they’re easy to understand.
```
gildedgardenia: I would just end the sentence at "worry". If a student is struggling to understand CSVs, seeing "they're easy to understand" could be disheartening.

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+CSV files are conceptually similar to simple spreadsheets in that they
+organize data in rows and columns. The major difference is that they
+don’t have the ability to carry out calculations the way spreadsheets
+do, and you can easily open, read, and edit them in plain text editors.
```
gildedgardenia: Anthropomorphizing spreadsheets and then addressing the reader in the same paragraph was a bit confusing. Perhaps instead:
"CSV files are conceptually similar to simple spreadsheets in that data is organized in rows and columns. The major differences are that we can't carry out calculations in a CSV the way that we can with a spreadsheet, but we *can* easily open, read, and edit CSV files in plain text editors."

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+corresponds to a new line. So each line below the first will constitute
+a row of data, or a record.
+
+Have a look at the data below line 1, and ask yourself the following
+questions:
+
+#. Which fields match up with which column names above?
+#. Why do some lines/rows (e.g. line 10) have more commas than others, if
+   commas are supposed to separate columns?
+#. What role do the double-quotes play?
```
gildedgardenia: I would maybe give a line number with the double quotes on them. It seems minor, but when I was looking through the file I couldn't find any. I had to use the search function to locate some.

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+It begins by declaring two local variables: ``columnChoices`` and
+``actionChoices``. These contain information relating to the menus that
+we’ll display, and we’ll look at them in more detail later.
+
+Next, we notice a ``while loop`` that starts ``while (true)``. While we usually
+want to avoid creating infinite loops, we have a good reason for doing so in
+this case! We want our application to continually run until the user decides
+they want to quit. The simplest way to do this is to loop forever. When the
+user wants to quit, they can kill our program by pressing *ctrl+C* (a
+widely-known command to kill a console application).
```
gildedgardenia, line 162: A student might ponder why use an infinite loop and force quit with *ctrl+C* when a ``do...while`` loop could do something similar. It might be worth addressing!

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+
+The ``GetUserSelection`` method takes in a string to display above the
+menu, to provide context for what the user is being asked. It also takes in
+a dictionary with string keys and string values. How is this used? What
+will this dictionary contain when the method runs?
+
+To figure this out, right-click on the method name and select *Find (All)
+References*. This will open a pane and display each location in the program
+where ``GetUserSelection`` is invoked. The first such usage is the first
+line of the main ``while loop``:
```
gildedgardenia: Maybe instead of "the main ``while loop``", "the ``while`` loop in the ``Main`` method."

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+
+.. sourcecode:: c#
+
+   string actionChoice = GetUserSelection("View jobs by:", actionChoices);
+
+What is this dictionary named ``actionChoices``? If we look a few lines
+above, we see:
```
gildedgardenia, line 199: Since the two are so far apart, maybe "If we look up at line 13, we see:"

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+provide an ordering to ``choices``, which doesn’t have an ordering of
+its own.
+
+The ``JobData`` Class
+---------------------
+
+The ``JobData`` class is responsible for importing the data from the CSV
+file and parsing it into a C#-friendly format; that is, into
+``Dictionary`` and ``List`` form. Look toward the bottom of the class
+and you will see a method named ``LoadData``, which does just what it
```
gildedgardenia, line 267: I would just give them a line number.

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+advertises. After parsing the file data, it stores the data in the
+private property ``AllJobs`` which is of type
+``List<Dictionary<string, string>>``.
+
+.. admonition:: Note
+
+   We haven’t covered static properties and methods in-depth yet. For this
+   assignment, know simply that they allow us to use properties and methods
+   of a class without creating an object from that class. For example, we
+   can call ``JobData.FindAll()`` from the ``Program`` class.
```
gildedgardenia: Maybe, "We haven't covered static properties and methods in-depth yet. For this assignment, we just need to know that using the ``static`` keyword with properties and methods allows us to use those properties and methods outside of the class *without* creating an object from that class. For example, we can call ``JobData.FindAll()`` from the ``Program`` class."

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+
+   If you want to create a new method in ``JobData``, or add a property, be
```
gildedgardenia: "If you want to create a new method or add a property in ``JobData``, be..."

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+   dictionary key. If a new field is added to the job records, this approach
+   will print out the new field without any updates to ``PrintJobs``.
+
+Test this method before moving on to your next step:
+
+#. Run your program with your changes. Select “1” to list the jobs, and then “0” to list them all.
+#. Make sure the printout matches the styling above.
+#. To test what happens if no jobs are found, select
+   “0” to search and then “3” to search for a location. Then, enter a location
+   that is not in the data (e.g. “Cancun”). Your message should be displayed.
```
gildedgardenia: "Your error message"

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+
+Create Method ``FindByValue``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+At this stage, the application will allow users to search a *given
+column* of the data for a given string. Your next task is to enable a
+search that looks for the search term in *all* of the columns.
+
+In the ``JobData`` class, create a new (``public static``) method that
```
gildedgardenia: The parentheses here are unnecessary.

src/assignments/tech-jobs-console.rst
```diff
@@ -0,0 +1,492 @@
...
+#. Prints each field of a job when using search functionality, and when
+   listing all columns. If there are no search results, a descriptive
+   message is displayed.
+#. Allows the user to search for a string across all columns.
+#. Returns case-insensitive results.
+
+Solution Demo
+-------------
+
+Watch a demo of a working solution.
```
gildedgardenia, line 469: I would add a note that the solution is in Java so that it doesn't confuse students.

gildedgardenia (commented):
I like the assignment a lot! Upon another read, I think I see what TAs and Impact meant by the assignments being too easy. So much of the starter code is explained within the page that the page is almost overwhelming to read at first glance and I didn't feel the need to look anything up outside of the page. Just food for thought!

src/assignments/tech-jobs-console.rst
CarlyLanglois: it didnt happen to me on my mac or the windows - was it just that one time or more than once? or each time you quit?

src/assignments/tech-jobs-console.rst
CarlyLanglois: i'm not sure i see how a do-while would benefit this solution or provide something substantially different? doesnt it still create the same situation?

src/assignments/tech-jobs-console.rst
CarlyLanglois: i just added a line number onto the action choice definition to add context

src/assignments/tech-jobs-console.rst
CarlyLanglois: let's keep it to encourage them to search for it themselves

src/assignments/tech-jobs-console.rst
CarlyLanglois: i changed it to "no results"

src/assignments/tech-jobs-console.rst
CarlyLanglois: huh? this video is for the c# solutions. do you see something different?

CarlyLanglois:
pushed up changes you recommended with a few outstanding. that's an interesting note - i hadn't considered that there was too much code explanation at the top, except that it took me a long time to initially read. for what its worth though, i do remember this assignment being tricky for me as a ta a few years ago ... it's worth taking a look but i'd argue we should put a pin in updating it for right now for the sake of timing - and knowing that updates here should also be applied to the java assignments

src/assignments/tech-jobs-console.rst
gildedgardenia: Turns out my computer was being weird!

src/assignments/tech-jobs-console.rst
gildedgardenia: I clicked on the wrong vid on Youtube. My bad!

src/assignments/tech-jobs-console.rst
gildedgardenia: Then I would change the region that you point them to because ``LoadData`` is pretty much in the middle. I would either say, "Look toward the middle of the class" or "If you scroll down the page, you will see a method named ``LoadData`` ..."

src/assignments/tech-jobs-console.rst
gildedgardenia: It would! I can just see students asking why the starter code doesn't use one! If you are not worried about it, then no need to add any additional explanation.

src/assignments/tech-jobs-console.rst
gildedgardenia: Awesome! Thanks!

gildedgardenia (approved):
Approved! I would add a note to the top that explains that this assignment explains a lot of the starter code, but that is not always going to be the case for all of the assignments.

PR #14, opened 2020-02-23
Lsn3 content

CarlyLanglois:


src/chapters/classes/modifiers.rst
```diff
@@ -0,0 +1,161 @@
...
+We should avoid giving everything default access when creating classes in C#
+and instead think carefully about what level of access each field and method
+should have.
+
+The table below details whether or not information can be accessed at different
+levels based on the access modifier. For example, a field with the ``private``
+access modifier can be accessed within the class, but cannot be accessed
+outside the class at the assembly or world-level. In C#, an **assembly** refers
+to a grouping of classes and other resources that form a particular unit of an application.
+**World-level** is the  level of the whole application and contains all of the packages and
```
gildedgardenia: Two spaces between "the" and "level".

src/chapters/classes/encapsulation.rst
```diff
@@ -0,0 +1,289 @@
...
+
+.. sourcecode:: c#
+
+   int numberOfCredits = 0;
+   double gpa = 0.0;
+
+.. index:: ! instance variable
+
+Fields are also referred to as **instance variables**, since they belong
+to an instance of a class.
```
gildedgardenia: I am having a hard time with this definition. I associate a field with a class variable, which is very separate from an instance variable. I think this sentence is oversimplifying the difference between the two.

src/chapters/classes/encapsulation.rst
```diff
@@ -0,0 +1,289 @@
...
+
+.. index:: ! accessor, ! getter, ! setter, ! auto-implemented property, ! backing field
+
+Getters and Setters
+^^^^^^^^^^^^^^^^^^^
+
+As declared, our four fields are private, which means that they are inaccessible 
+to code outside of the ``Student`` class. As a
+rule-of-thumb, *fields should always be private unless you have a very,
+very, very good reason to not make them so.* As we mention on the previous page, it is
```
gildedgardenia: I think the bit about fields needing to be private should be in the above section. Then you can transition with, "As stated above, fields should be private. So how do we then access the value of a field or change that value? We use methods called getters and setters." Or something like that. It might also transition better between the two sections that way.

src/chapters/classes/encapsulation.rst
```diff
@@ -0,0 +1,289 @@
...
+       private int studentId;
+       private int numberOfCredits = 0;
+       private double gpa = 0.0;
+   }
+
+In order to provide access to private fields, **getter and setter methods** 
+are used. Getters and setters do what you might guess: get and
+set a given field. If we make the getter and/or setter method for a
+given property public, then others will be able to access or modify the
+field in that way.
```
gildedgardenia, line 100: So I made the above comment before reading these lines. I think these lines should start this section and then the above lines could serve as the final info about fields in the fields section.

src/chapters/classes/encapsulation.rst
```diff
@@ -0,0 +1,289 @@
...
+private if you’re just going to allow people to get and set them
+anyway!?” Great question. There are lots of reasons to use getters and
+setters to control access. Here are just a few:
+
+1. Sometimes, you’ll want to implement behavior that happens every time a
+   field is accessed (get) or changed (set). Even if you can’t think of
+   such a reason when writing your class, you might later have the need
+   to add such behavior. If you don’t use getters and setters, you’ll
+   have to do a lot more refactoring if you ever decide to add such
+   behaviors.
```
gildedgardenia: I had to read this item a couple of times through to get the point. Maybe rephrase?

gildedgardenia (commented):
The chapter is looking good so far! I highlighted a few places where I thought that a rephrasing could maybe clarify the point further and one small typo.

src/chapters/classes/encapsulation.rst
CarlyLanglois: good call! im just going to remove this since we haven't introduced the instance/static distinctions yet

src/chapters/classes/encapsulation.rst
CarlyLanglois: I changed some of this up more along the lines of this comment than the next one. I like the idea to put this above the section header to help with the transition rather than adding it to the end of the section because a) we need an explanation for why we use getters/settters and thats harder without already discussing why fields should be private. and b) this section is long and the end is closer to the auto-implemented property which draws upon a more implicit private field which could be confusing imo

src/chapters/classes/encapsulation.rst
CarlyLanglois: see my note above ^^

gildedgardenia (approved):
Awesome! I love the changes you made!

PR #15, opened 2020-02-24
Lesson 3 Exercises

gildedgardenia:


src/chapters/classes/exercises.rst
```diff
@@ -0,0 +1,27 @@
+.. _classes-exercises-part1:
+
+Exercises: Classes and Objects
+==============================
+
+Before you get started, make sure you have forked and cloned the `starter code <https://github.com/LaunchCodeEducation/csharp-web-dev-lsn3and4classes>`_.
+
+#. Open up the file, ``Student.cs``, and add all of the necessary getters and
+   setters. Think about the access level each field and method should have, and
+   try reducing the access level of at least one setter to less than public.
```
CarlyLanglois: since the starter code gives this:

public string Name { get; set; } ...

the fields are already set and private - plus they have getters and setters. what about replacing this item to either have the students write the auto-implementing properties themselves? or starting with ``private name ...`` and asking them to update the fields to the auot-implemented properties?

src/chapters/classes/exercises.rst
```diff
@@ -0,0 +1,27 @@
...
+#. Instantiate the ``Student`` class using yourself as the student. For the
```
CarlyLanglois: we might also want to add "In Program.cs" to the beginning of this item to be clear what it is for

src/chapters/classes/exercises.rst
```diff
@@ -0,0 +1,27 @@
...
+#. In the ``SchoolPractice`` project, create a class ``Teacher`` with four fields:
+   ``firstName``, ``lastName``, ``subject``, and ``yearsTeaching``. Add getters
+   and setters to the class and add the access level to each field and method
+   in the class. When adding your getters and setters, think about what we read in the section on encapsulation.
+
+   a. What access modifier restricts access the most for what we need?
+   b. If it is a field, could we restrict the access to ``private`` and use
+      getters and setters?
+   c. If we do set fields to ``private``, what access level do we have to give
+      our getters and setters?
```
CarlyLanglois: for this item too, i'd argue that we want the end result to be these auto-implemented properties which could influence how we tell the students to create those fields

CarlyLanglois (commented):
looks great basically but the auto-implemented properties i think are different enough from creating java fields and getters/setters that we should update these instructions for this c# version.
maybe the way to go is more of a two step approach where we first ask the students to write the fields with a certain access level and then direct them to update them using this auto-implementing action

src/chapters/classes/exercises.rst
gildedgardenia: Done!

src/chapters/classes/exercises.rst
gildedgardenia: I went with the second option!

src/chapters/classes/exercises.rst
gildedgardenia: Got it!

CarlyLanglois (approved):
looks great! just one thing for the starter code repo - the field names should be lowercased.
i know you wont but don't forget to merge in master!

PR #24, opened 2020-03-09
Graded Assignment #2

gildedgardenia:


src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+#. Use TDD to design and code a custom ``ToString`` method for the ``Job``
+   class.
+#. Use inheritance to DRY the code within ``Employer``, ``Location``,
+   ``CoreCompetency``, and ``PositionType``.
+
+Explore the ``Employer`` Class
+------------------------------
+
+Open the ``Employer`` file in Visual Studio and examine the code. In addition to the
+two fields---``Id`` and ``Value``---the class includes some custom methods like ``ToString`` and ``equals``.
```
CarlyLanglois: the fields are lowercased, properties are uppercase. i think its important to emphasize the difference, but do you feel differently? I feel like i keep writing this and if you have a different opinion, please let me know.

CarlyLanglois: and the method is Equals

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+
+         // Additional methods omitted from this code block
+      }
+
+#. Line 3 declares the variable ``nextId``. Since it is ``static``, its
+   changing value is NOT stored within any ``Employer`` object.
+#. The first constructor (lines 6 - 10) accepts no arguments and assigns the
+   value of ``nextId`` to the ``Id`` field. It then increments ``nextId``.
+   Thus, every new ``Employer`` object will get a different ID number.
+#. The second constructor (lines 12 - 15) assigns ``value`` to the ``Value``
```
CarlyLanglois: can you say this differently because the field is `value` so correcting this to assigns `value` to the `value` field is not particularly helpful.

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+   field. However, it ALSO initializes ``Id`` for the object by calling the
+   first constructor statement with the `` : this()`` syntax. Including ``this()`` in
```
CarlyLanglois: this syntax highlighting isnt taking - maybe because there is a space inside?

CarlyLanglois: ``: this()``

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+
+.. admonition:: Tip
+
+   Save, commit, and push your work to GitHub.
+
+Use Unit Testing to Verify Parts of the ``Job`` Class
+-----------------------------------------------------
+
+Instead of manually creating sample ``Job`` objects to verify that your class
+works correctly, you will use unit tests instead.
```
CarlyLanglois: i'd just remove this line - i don't think we need to assume that this would be a default behavior

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+``Name``, ``Employer``, ``EmployerLocation``, ``JobType``, ``JobCoreCompetency``.
+There’s also an ``Id`` field which will be used to uniquely identify ``Job``
+objects.
+
+The main difference between the object representation of a job and the
+string-based representation is that the values of ``EmployerName``, ``EmployerLocation``,
+and the other non-ID fields are no longer strings. Instead, they are classes of
+their own.
+
+Job Fields
```
CarlyLanglois: can we call this job members instead? since we're handling both the properties and the backing fields?

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+the other properties. These classes---``Employer``, ``Location``,
+``CoreCompetency``, ``PositionType``---have ``Value`` and ``Id`` fields.
+
+So, for example, if you had a ``Job`` instance, you could get the name of the
+employer this way:
+
+.. sourcecode:: csharp
+
+   // job is an instance of Job
+   string employerName = job.Employer.Value;
```
CarlyLanglois: job doesn't have an Employer property as the starter is written - although if you're interested in making this assignment a bit more object-oriented, i think combining employer and location is not a bad idea. having separate classes for each of these items that just only carry a unique value feels a little shallow so combining two of them gives us a slightly more realistic design imo. that is, if we decide to buy into the idea that a given employer has only one location for this assignment

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+
+Additionally, the ``ToString()`` method of the ``Employer`` class is set up to
```
CarlyLanglois, line 108: i love writing methods with the parentheses - its a nice visual reminder of kind of entity we refer to - and its easier said than done but let's try to be consistent within the page if we use them or not

src/assignments/tech-jobs-oo.rst
CarlyLanglois: also to this point, the explanation that follows about why we want to make these fields their own classes makes more sense when at least one of the classes does more, or carries more information, than just an id and value

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+#. It should call the first constructor to initialize the ``Id`` field.
+#. It must also initialize the ``Value`` field for a new ``Location`` object.
+
+.. _generator-shortcut:
+
+The ``CoreCompetency`` Class
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Open the class file. In this case, the constructors and custom methods are
+ready. Sally needs you to change the ``Id`` and ``Value`` fields to auto-implemented properties, but NOT ``nextId``.
```
CarlyLanglois: great idea to ask students to add auto-implemented props - if we're going to give them some starter code for the fields before their properties are implemented, lets make the fields lowercased

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+
+   Save, commit, and push your work to GitHub.
+
+Use Unit Testing to Verify Parts of the ``Job`` Class
+-----------------------------------------------------
+
+Instead of manually creating sample ``Job`` objects to verify that your class
+works correctly, you will use unit tests instead.
+
+Create a new project inside the ``TechJobsOO`` solution called ``TechJobsTests``, then
```
CarlyLanglois, line 302: i'm in favor of not getting specific about the other setup steps involved to get an mstest project wired up to a project whose class it is testing, but i do think its a good idea to give some indication of the fact that connecting the two projects is required. what do you think?

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+sequential integers.
+
+#. In ``JobTests``, define a test called ``TestSettingJobId``.
+#. Create two ``Job`` objects using the empty constructor.
+
+   .. admonition:: Note
+
+      Instead of creating the ``Job`` objects inside the test method, you could use ``[TestInitialize]``.
+
+#. Use ``Assert.AreEqual``, ``Assert.IsTrue``, or ``Assert.IsFalse`` to test that the
```
CarlyLanglois: can i suggest combining these two steps and removing the note about testinitialize? I'd get rid of that suggestion here because it doesn't really start to add benefit to a test class until more than one test is written. and combining these two steps gives me a better idea of why we're asking the reader to initialize two job objects

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+these fields are ``int``, ``string``, ``Employer``, ``Location``,
+``PositionType``, and ``CoreCompetency``, respectively.
+
+#. In ``JobTest``, define a test called
+   ``TestJobConstructorSetsAllFields``.
+#. Declare and initialize a new ``Job`` object with the following data:
+
+   .. sourcecode:: csharp
+
+      new Job("Product tester", new Employer("ACME"), new Location("Desert"), new PositionType("Quality control"), new CoreCompetency("Persistence"));
```
CarlyLanglois: i see two candidates for dialing back the level of instruction we're giving here - a) we can consider removing this intro language describing what type each Job member is - this should be pretty well grasped by most.
and b) instead of providing the whole right side of the sample job instance as it will be programmed, we could replace this sourcecode directive with a description of what we want each field to be

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+   class and value of each field.
+
+   .. admonition:: Tip
+
+      The ``is`` keyword can be used to check the class of an object.
+      The result of the comparison is a boolean.
+
+      .. sourcecode:: csharp
+
+         objectName is ClassName VariableName
```
CarlyLanglois: this is a good opportunity to reference the other time we're used "is" in the lsn 4 content - and also i don't necessarily read this definition and the other use of "is" as consistent - arent you saying in the other use case that whats results is a conversion of type rather than a bool?

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+#. In ``JobTest``, define a test called ``TestJobsForEquality``.
+#. Generate two ``Job`` objects that have identical field values EXCEPT for
+   ``Id``. Test that ``Equals`` returns ``false``.
+
+It might seem logical to follow up the ``false`` case by testing to make sure
+that ``Equals`` returns ``true`` when two objects have the same ID. However,
+the positive test is irrelevant in this case.
+
+The way you built your ``Job`` class, each ``Id`` field gets assigned a unique
+value, and the class does not contain a setter for the ``Id`` field. You also verified
```
CarlyLanglois: does it not?

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+
+#. Modify line 5 to *extend* the ``JobField`` class into ``Employer``.
+
+   .. sourcecode:: csharp
+      :lineno-start: 5
+
+      public class Employer : JobField
+      {
+         //Code not displayed.
+      }
```
CarlyLanglois: here's another directive id consider removing to increase the challenge

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+
+#. Next, remove any code in ``Employer`` that matches code from ``JobField``
+   (e.g. the ``Id``, ``Value``, and ``nextId`` fields are shared).
+#. Remove any of the custom methods that are identical.
```
CarlyLanglois: none of these classes really have any custom methods, do they? im not looking at the solution code

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+   .. sourcecode:: csharp
+      :lineno-start: 7
+
+      public Employer(string value) : base() 
+      {
+         Value = value;
+      }
+
+   The ``:`` and ``base`` keywords link the ``JobField`` and
+   ``Employer`` classes.
```
CarlyLanglois: consider a stronger explanation of this syntax - or none at all since i presume you'll include this in the inheritance lesson content

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+   Employer: Data not available
+   Location: Home
+   Position Type: UX
+   Core Competency: Tasting ability
+
+Excellent! You successfully shifted the old console app into a more useful
+object oriented configuration.
+
+Now that the new structure is ready, another team member can refactor the
+import and display methods to use the new classes. Once these are ready, our
```
CarlyLanglois: there are no import and display methods, right? or am i missing them

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,580 @@
...
+
+Sanity Check
+-------------
+
+Once you finish all of the tasks outlined above, all that remains is to check
+the console display.
+
+Sally has provided some commented-out code in ``Main`` that prints out a small
+list of ``Job`` objects. Go ahead and activate this code and run it.
+Properly done, your output should look something like:
```
CarlyLanglois: fyi this is not in the current version of the starter code - also something we talked about with that studio repo, add a gitignore so we're not trekking around a vs settings file. 
also,  you may wnat to clarify you're referring to the main method in program, unless the aim is to have the students deduce for themselves that you are referring to a method in a class file and not a class name.

src/assignments/tech-jobs-oo.rst
gildedgardenia: Fixed!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Changed the parameter name to help!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Fixed!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Gotcha!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Done!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Fixed!

src/assignments/tech-jobs-oo.rst
gildedgardenia: I added a little something. Let me know your thoughts!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Combined the steps and removed the note!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Done!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Changed to just "methods" since they can add the methods like ``Equals()`` to the parent class.

src/assignments/tech-jobs-oo.rst
gildedgardenia: It is covered, so this should be an okay step!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Fixed the ``Employer``! That is supposed to be ``EmployerName``!

I do like the organization as is because it gives them something to think about when it comes to streamlining the organization at the end. I did add a little note to the end of the next section to indicate that the app could be better organized and that students will need to do some work to make it so!

src/assignments/tech-jobs-oo.rst
gildedgardenia: I think I got them all!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Done and done!

src/assignments/tech-jobs-oo.rst
gildedgardenia: It should not! Fixed!

src/assignments/tech-jobs-oo.rst
gildedgardenia: That is a reference to the work they did in Assignment #1. Clarified!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Fixed!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Fixed this explanation so it made more sense and referenced other content

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+So, for example, if you had a ``Job`` instance, you could get the name of the
+employer this way:
+
+.. sourcecode:: csharp
+
+   // job is an instance of Job
+   string employerName = job.EmployerName.Value;
+
+Additionally, the ``ToString()`` method of the ``Employer`` class is set up to
+return the ``Value`` property. Thus, using one of these objects in another string
```
CarlyLanglois: here's a doc page that might help clarify the field/property distinction.
in brief, I understand properties - as the term is used in the c# context - to behave like methods. Properties do the getting and setting actions, fields hold the values.
So rather than saying that a Value property is returned, it seems more appropriate and descriptive of the language to say that the value field is returned

CarlyLanglois: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+---------------
+
+The list below provides a general overview of your assigned tasks. Specific
+details for each part appear in the following sections, so be sure to read them
+carefully as you solve each problem.
+
+#. Review Sally's code in the ``Employer`` class to learn how to assign a
+   unique ID.
+#. Add getters, setters, and custom methods as needed to the ``Location``,
+   ``CoreCompetency``, and ``PositionType`` classes.
```
CarlyLanglois: check your starter code on this one - some of these classes already have the accessors and those that dont should have a lowercase field name. I dont know if you get these hints in your ide, but VS gives me a hint when i've accidentally named a field starting with an uppercase letter

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+
+         public Employer (string v) : this()
+         {
+            Value = v;
+         }
+
+         // Additional methods omitted from this code block
+      }
+
+#. Line 3 declares the variable ``nextId``. Since it is ``static``, its
```
CarlyLanglois: declares the field

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+   changing value is NOT stored within any ``Employer`` object.
+#. The first constructor (lines 6 - 10) accepts no arguments and assigns the
+   value of ``nextId`` to the ``Id`` field. It then increments ``nextId``.
```
CarlyLanglois: ``id`` field

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+   Thus, every new ``Employer`` object will get a different ID number.
+#. The second constructor (lines 12 - 15) assigns ``v`` to the ``Value``
+   property. However, it ALSO initializes ``Id`` for the object by calling the
+   first constructor statement with the ``: this()`` syntax. Including ``this()`` in
+   any ``Employer`` constructor makes initializing ``id`` a default behavior.
```
CarlyLanglois: see my note above about the field/property distinction.
couple other notes here, we don't need "however" in the second sentence. and since its new, i'd use `: this()` each time you refer to the syntax (ie, with the colon).
lastly, i don't necessarily think that that changing the constructor's parameter to 'v' helps to clarify what this accomplishes. you can keep the parameter name the same (it is still called value in the starter code anyway) and just adjust the explanation - "the second constructor assigns the value field"

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+``Employer`` class for hints on how to structure your code.
+
+The ``Location`` Class
+^^^^^^^^^^^^^^^^^^^^^^
+
+Open the ``Location.cs`` file. Note that the
+methods for this class are done, as is the constructor for initializing the
+``Id`` property.
+
+Sally left you a ``TODO`` comment with instructions for coding a second
```
CarlyLanglois, line 248: can you update this todo in the starter code with the field/property distinction

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+constructor:
+
+#. It should call the first constructor to initialize the ``Id`` property.
+#. It must also initialize the ``Value`` property for a new ``Location`` object.
+
+.. _generator-shortcut:
+
+The ``CoreCompetency`` Class
```
CarlyLanglois, line 256: the starter code uses the this dot syntax but this can be truncated to just the property name

src/assignments/tech-jobs-oo.rst
CarlyLanglois: yeah this good - maybe even one step further?
Add the appropriate dependency to ``TechJobsTests`` to test the classes in the ``TechJobsOO`` project.
^^ that gives them a bit of a hint of what we mean by the appropriate dependency

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Each ``Job`` object should contain six properties---``Id``, ``Name``, ``EmployerName``,
+``EmployerLocation``, ``JobType``, and ``JobCoreCompetency``.
+
+#. In ``JobTest``, define a test called
+   ``TestJobConstructorSetsAllFields``.
+#. Declare and initialize a new ``Job`` object with the following data: ``"Product tester"`` for ``Name``, ``"ACME"`` for ``EmployerName``, ``"Desert"`` for ``JobLocation``, ``"Quality control"`` for ``JobType``, and ``"Persistence"`` for ``JobCoreCompetency``.
+#. Use ``Assert`` statements to test that the constructor correctly assigns the
+   class and value of each field.
```
CarlyLanglois: i don't think we need to tell them to check the class of each field? i would just ask to check that the field values are assigned

CarlyLanglois: for that matter, looking at your solution now, if this is why the tip about using "is" below is included, i don't think we need this. there's no need to test that the job objects created in the test file are job objects since that is what they are instantiated as. instead, the test for this step should just focus on the field assignments - ie, testJob.EmployerName.Value == "whatever"

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+      .. sourcecode:: csharp
+
+         objectName is ClassName VariableName
+
+Test the ``Equals()`` Method
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Two ``Job`` objects are considered equal if they have the same ``id`` value,
+even if one or more of the other fields differ. Similarly, the two objects
+are NOT equal if their ``Id`` values differ, even if all the other fields are
```
CarlyLanglois: id values

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+      Location: _______
+      Position Type: _______
+      Core Competency: _______
+
+#. If a field is empty, the method should add, "Data not available" after
+   the label.
+#. (Bonus) If a ``Job`` object ONLY contains data for the ``Id`` field, the
+   method should return, "OOPS! This job does not seem to exist."
+
+In ``JobTests``, add a new test to check the first requirement, then run
```
CarlyLanglois: first requirement (item 1. above) or something like this

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+#. Consider the following questions to help you decide what code to put in the
+   ``JobField`` class:
+
+   a. What fields do ALL FOUR of the classes have in common?
+   b. Which constructors are the same in ALL FOUR classes?
+   c. Which custom methods are identical in ALL of the classes?
+
+#. In ``JobField``, declare each of the common fields.
+#. Code the constructors.
+#. Add in the custom methods.
```
CarlyLanglois: what about, "Add in any inherited method overrides"

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+Extend ``JobField`` into ``Employer``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Now that you have the common code located in the ``JobField`` file, we can
+modify the other classes to reference this shared code. Let's begin with
+``Employer``.
+
+#. Modify line 5 to *extend* the ``JobField`` class into ``Employer``.
+#. Next, remove any code in ``Employer`` that matches code from ``JobField``
+   (e.g. the ``Id``, ``Value``, and ``nextId`` fields are shared).
```
CarlyLanglois: fields are lowercased

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+   .. sourcecode:: csharp
+      :lineno-start: 7
+
+      public Employer(string value) : base() 
+      {
+         Value = value;
+      }
+
+   The ``:`` and ``base`` keywords link the ``JobField`` and
+   ``Employer`` classes.
```
CarlyLanglois: what i mean here is that this sentence is a little black-boxy - it doesn't really explain what this syntax is -and kind of obscures what ": base" is used for.  Since this will be covered in the inheritance lesson content, what are we adding by keeping this? if you want to keep some explanation of the fact that we need to explicitly inherit the no-arg constructor for the parent class, we should be more detailed in explaining that here. 
describing `:base` as keywords that link two classes is not specific enough to be meaningful

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+
+Now that you have the common code located in the ``JobField`` file, we can
+modify the other classes to reference this shared code. Let's begin with
+``Employer``.
+
+#. Modify line 5 to *extend* the ``JobField`` class into ``Employer``.
+#. Next, remove any code in ``Employer`` that matches code from ``JobField``
+   (e.g. the ``Id``, ``Value``, and ``nextId`` fields are shared).
+#. Remove any of the methods that are identical.
+#. The empty constructor is shared, but not the second. Replace the two
```
CarlyLanglois, line 475: actually, i don't think we need a second constructor in JobField, do we? i don't think it is used anywhere

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+   You know you need to do this, but here is the reminder anyway. Save, commit,
+   and push your work to GitHub.
+
+Sanity Check
+-------------
+
+Once you finish all of the tasks outlined above, all that remains is to check
+the console display.
+
+Sally has provided some commented-out code in ``Program.cs`` that prints out a small
```
CarlyLanglois, line 505: its funny because this time you actually have provided it :)

src/assignments/tech-jobs-oo.rst
```diff
@@ -0,0 +1,558 @@
...
+Now that the new structure is ready, another team member can refactor the
+import and display methods from the previous assignment to use the new classes. Once these are ready, our
+team will refine the search features and move the app online to provide a
+better user interface.
+
+How to Submit
+--------------
+
+To turn in your assignment and get credit, follow the
+:ref:`submission instructions <how-to-submit-work>`.
```
CarlyLanglois, line 546: on this submission instructions page, add a link to this assignment too

CarlyLanglois (commented):
looks great, sorry to leave one more "hey check all of your instances of *blank*" type note, but i think its important we deliver a consistent message about what fields are and what properties are in the context of c#. I left a note in line somewhere but adding it here to flag it, when we refer to a class property in c#, we're essentially taking about a behavior, vs. a field which is a variable for some class data. i think this page could help define the difference: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties

and believe me, im still sussing out the difference myself so im sure im guilty of conflating the two as well

src/assignments/tech-jobs-oo.rst
gildedgardenia: I changed up this goal a bit to better account for the starter code.

src/assignments/tech-jobs-oo.rst
gildedgardenia: Done!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Done!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Done!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Done!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Fixed!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Fixed!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Done!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Done!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Fixed!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Ok!

src/assignments/tech-jobs-oo.rst
gildedgardenia: I removed the sentence!

src/assignments/tech-jobs-oo.rst
gildedgardenia: We do need to use that second constructor. The solution is not properly using it. Fixing!

src/assignments/tech-jobs-oo.rst
gildedgardenia: Gotcha!

CarlyLanglois (commented):
i think this is in good shape - 
do you want me to look at your updated soln code? i don't think its pushed up yet.
also heads up - the starter code relies on some properties that arent fully implemented until a couple of steps in which could cause some confusion. you might want to consider commenting those out to begin - or ask the students to implement the code that relies on them.

CarlyLanglois (approved):

PR #26, opened 2020-03-11
Lsn6 studio

CarlyLanglois:


src/chapters/inheritance/studio.rst
```diff
@@ -0,0 +1,85 @@
+.. _inheritance-studio:
+
+Studio: Fun with Quizzes
+========================
+
+For this studio, you will design and build a console program that allows
+the user to take a quiz. This means you will have to create some
+questions, and get some input from the user.
```
gildedgardenia: This comma is unnecessary.

src/chapters/inheritance/studio.rst
```diff
@@ -0,0 +1,85 @@
...
+
+First, the questions. We want to be able to handle multiple types of
+questions:
+
+#. **Multiple choice**: a question with a fixed set of possible answers, of which only one may be chosen 
+   and only one answer is correct
+#. **Checkbox**: a question with a fixed set of possible answers, of which any number may be chosen; there is 
+   one correct combination of choices
```
gildedgardenia: I would break down these long sentences into multiple sentences. It is a little difficult to figure out what is going on.

gildedgardenia (commented):
I like it! I think the wording is confusing between the ``Question`` class and the "question classes". Is there another way to refer to the "question classes" to help alleviate this confusion?

gildedgardenia (approved):
Looks great!

PR #28, opened 2020-03-16
Lsn7

CarlyLanglois:


src/chapters/interfaces-and-polymorphism/introduction.rst
```diff
@@ -0,0 +1,64 @@
+Introduction
+============
+
+.. index:: ! polymorphism
+
+The third pillar of Object-oriented programming that we’ll
```
gildedgardenia: I think "object" can be lower-cased.

src/chapters/interfaces-and-polymorphism/interfaces.rst
```diff
@@ -0,0 +1,294 @@
+Interfaces
+==========
+
+.. index:: ! interface
+
+An **interface** is a formal construction within C#. Interfaces allow us to
+create a set of behaviors that a class must fulfill. A C# interface may
+contain:
```
gildedgardenia: I might add a note here that interfaces exist in other languages too.

src/chapters/interfaces-and-polymorphism/interfaces.rst
```diff
@@ -0,0 +1,294 @@
...
+^^^^^^^^^^^^^^^^^
+
+.. index:: ! method signature
+
+One really useful aspect of using an interface is the ability to specify method signatures.
+A **method signature** includes the name, parameters, and return
+type of a method, but no body.
+
+As part of our cat application, let's create a method signature, ``Eat``, as part of an 
+interface, ``IFeedable``.
```
gildedgardenia: I might add a note that interfaces in C# start with "I" as a naming convention.

gildedgardenia: I found it! I might move that note up earlier!

src/chapters/interfaces-and-polymorphism/interfaces.rst
```diff
@@ -0,0 +1,294 @@
...
+released, while not forcing those already using the interface to add new
+code to their classes. *You should avoid using default methods in all situations other than the 
+one described here.*
+
+Implementing an Interface
+-------------------------
+
+The purpose of an interface is to define a contract of behaviors that classes uphold. In doing so, we say that they “*implement* the
+interface”. The syntax for implementation is the same as that for
+inheritance. Here’s how we can use the ``IFeedable`` interface in
```
gildedgardenia: Good spot to remind them why the naming convention is so important!

src/chapters/interfaces-and-polymorphism/interfaces.rst
```diff
@@ -0,0 +1,294 @@
...
+         }
+
+         // ...rest of the class definition...
+
+      }
+
+Since we’ve declared that ``Cat`` implements ``IFeedable``, we have to
+provide an implementation for the ``Eat`` method, with the signature as
+specified in the interface definition. Note that we do not use ``override``
+since the class is *implementing* the method, rather than overriding it.
```
gildedgardenia: It might be worth noting earlier that these methods do not need the ``virtual`` keyword to be overridden.

src/chapters/interfaces-and-polymorphism/interfaces.rst
```diff
@@ -0,0 +1,294 @@
...
+``CatSitter`` object.
+
+As is the case with classes inherited from others, interfaces also enable polymorphic usage of 
+objects. We can create an object and then use it in different contexts based on the
+interfaces that it implements.
+
+Similar to abstract classes, *interfaces may not be instantiated*.
+You may implement an interface, or declare variables and parameters as
+interface types. You cannot, however, create an instance of an
+interface.
```
gildedgardenia: I think this should be brought up way earlier.

src/chapters/interfaces-and-polymorphism/interfaces.rst
```diff
@@ -0,0 +1,294 @@
...
+Here are a few benefits of using interfaces:
+
+#. You can only extend one class, but you may implement many interfaces.
+#. You can extend a class and implement an interface at the same time.
+#. By declaring variables and parameters as interface types, you make
+   your code useful for a much wider variety of situations.
+#. When you declare properties and return types to be interface types,
+   you decouple code using your classes from the actual class types you
+   use. This means that you are free to change the specific
+   implementation of your classes without affecting those using them.
```
gildedgardenia, line 268: Awesome!

src/chapters/interfaces-and-polymorphism/interfaces-and-abstract-classes.rst
```diff
@@ -0,0 +1,41 @@
...
+The main differences between interfaces and abstract classes are:
+
+#. You *implement* an interface, while you *extend* an abstract class. The net effect of this is 
+   that a class may implement interfaces while also extending a class. Note that while you can 
+   implement *more than one* interface, you can only extend *one* class.
+#. Abstract classes may contain non-constant fields, while interfaces can only contain constant 
+   fields.
+#. Abstract classes should be used to collect and specify behavior by related classes, while an 
+   interface should be used to specify related behaviors that may be common across unrelated 
+   classes.
```
gildedgardenia: Can you elaborate on this point or maybe rephrase it? I think this final point is the one that is really driving home why we choose to use interfaces vs. abstract classes and it might be easy to miss the "related" vs. "unrelated". Maybe emphasizing those two words would be a simple fix that would help.

src/chapters/interfaces-and-polymorphism/interfaces-and-abstract-classes.rst
```diff
@@ -0,0 +1,41 @@
...
+
+.. admonition:: Example
+
+   Let's go back to our ``IFeedable`` interface. If we want to add a ``Dog`` class to our 
+   application, we might implement the ``IFeedable`` interface for our ``Dog`` class.
+   This makes sense as dogs are creatures that we feed. However, as dogs and cats are so 
+   different, it is unlikely that they would share many characteristics through a ``Pet`` class.
```
gildedgardenia: Maybe merge this example into the final point up above?

src/chapters/interfaces-and-polymorphism/given-interfaces.rst
```diff
@@ -0,0 +1,114 @@
+Interfaces In The Wild
+======================
+
+The first situations where you’ll want to use interfaces involve applying pre-defined interfaces 
+and classes that are part of C#. Here are a few examples.
+
+IComparer<T>
+------------
```
gildedgardenia, line 11: Do you have an example for this one?

src/chapters/interfaces-and-polymorphism/given-interfaces.rst
```diff
@@ -0,0 +1,114 @@
...
+`List <https://msdn.microsoft.com/en-us/library/6sh2ey19(v=vs.110).aspx>`__
+to sort its contents with the
+`Sort <https://msdn.microsoft.com/en-us/library/234b841s(v=vs.110).aspx>`__
+method.
+
+IEnumerable<T>
+--------------
+
+**Purpose**: Enable iteration over a collection of objects using
+``foreach``.
```
gildedgardenia, line 65: I think it would be helpful here and elsewhere on this page to acknowledge that students may have already done something similar (in this case, they have already used ``foreach``) and why interfaces make it the same/different? Are these interfaces that students were accessing this whole time? If so, how did they access these interfaces? I think some students will need additional clarity.

gildedgardenia (commented):
I didn't really notice any typos, so all of my comments are more focused on phrasing and clarity. I think this is another chapter where students either get it right away or they are left wondering why we even have to use the concepts the chapter covers.

src/chapters/interfaces-and-polymorphism/interfaces.rst
CarlyLanglois: i dont understand this note, can you explain what you mean here?

CarlyLanglois: ok fixed this from your note on slack. lmk what you think

src/chapters/interfaces-and-polymorphism/introduction.rst
```diff
@@ -0,0 +1,64 @@
+Introduction
+============
+
+.. index:: ! polymorphism
+
+The third pillar of object-oriented programming that we’ll
+explore is polymorphism.
```
gildedgardenia: This should be the fourth pillar of oop.

src/chapters/interfaces-and-polymorphism/introduction.rst
```diff
@@ -0,0 +1,64 @@
...
+            // ...code to prepare the cat's meal...
+
+            Pet.Eat();
+         }
+      }
+
+   The method ``FeedTheCat`` uses the property ``Pet``, which is of type
+   ``Cat``, but since a ``HouseCat`` *is a* ``Cat`` via inheritance, it is
+   perfectly acceptable to use an instance of ``HouseCat`` to fill the
+   ``Pet`` property.
```
gildedgardenia: I think this could be broken down into two sentences. "... which is of type ``Cat``. Since a ``HouseCat`` ..."
Also, maybe change the end to "instance of ``HouseCat`` as the ``Pet`` property."

src/chapters/interfaces-and-polymorphism/introduction.rst
```diff
@@ -0,0 +1,64 @@
...
+
+      HouseCat suki = new HouseCat("Suki", 12);
+      CatSitter annie = new CatSitter(suki);
+
+      annie.FeedTheCat();
+
+   Similarly, ``FeedTheCat`` can accept ``Tiger`` instances as well. This
+   is because the only thing that the method requires is that the input
+   parameter has the methods defined within ``Cat``, and via inheritance,
+   both of the subclasses satisfy this requirement.
```
gildedgardenia: Replace "of the subclasses" with "``HouseCat`` and ``Cat``"

src/chapters/interfaces-and-polymorphism/interfaces.rst
```diff
@@ -0,0 +1,301 @@
+Interfaces
+==========
+
+.. index:: ! interface
+
+An **interface** is similar to an abstract class, with some important differences. Interfaces allow us to 
+create code organized by behavior, rather than static data. While some object-oriented languages encourage 
+the use of interfaces as a design concept, C# (among others) includes interfaces as a formal construction.
```
gildedgardenia: Could this sentence be brought down a reading level? I think some students might read it and wonder why a design concept is different from a formal construction. Perhaps, "While some object-oriented languages encourage using interfaces to improve software design, C# is a language that includes interfaces ..."

src/chapters/interfaces-and-polymorphism/interfaces.rst
```diff
@@ -0,0 +1,301 @@
...
+      }
+
+In the code above, notice the following:
+
+#. We need to use the ``interface`` keyword to define our interface, ``IFeedable``.
+#. ``Eat`` only has a signature. We are NOT allowed to provide a body for methods defined 
+   in interfaces.
+#. ``Eat`` also does not have an access modifier. Interface members are always ``public``, 
+   and while we may use the ``public`` modifier, it’s unnecessary. An interface method may NOT 
+   have an access modifier that is more restrictive than ``public``.
```
gildedgardenia: Perhaps, "While we may use the ``public`` modifier, interface members are always ``public`` and can NOT have an access modifier more restrictive than ``public``. This makes it unnecessary for us to add the ``public`` modifier."

src/chapters/interfaces-and-polymorphism/interfaces.rst
```diff
@@ -0,0 +1,301 @@
...
+Once you get used to interfaces, you’ll begin to think more abstractly about which 
+*behaviors* your code requires rather than which *classes* your code requires. This means
+you will start to “code to interfaces” (an OOP principle) instead of
+coding to classes, and your code will become more flexible and
+extensible.
+
+Here are a few benefits of using interfaces:
+
+#. You can only extend one class, but you may implement many interfaces.
+#. You can extend a class and implement an interface at the same time.
```
gildedgardenia, line 262: It might be helpful to mention somewhere that you have to put the class before the interface.

src/chapters/interfaces-and-polymorphism/testing-interfaces.rst
```diff
@@ -0,0 +1,59 @@
...
+      [TestClass]
+      public class CatTests
+      {
+         [TestMethod]
+         public void TestHouseCatImplementsEatMethod()
+         {
+            IFeedable test_cat = new HouseCat("test");
+            Assert.AreEqual("the feedable is eating", test_cat.Eat());
+         }
+      }
```
gildedgardenia: Could these code blocks be broken up with more explanation between? Statements like, "If our code works, then ``Assert.AreEqual("the feedable is eating", test_cat.Eat());`` evaluates to true." might be really helpful to reinforce the concepts.

gildedgardenia (approved):
It is looking really good! I found a few more points of confusion that are worth another look, but other than that this chapter is pretty much good to go!

src/chapters/interfaces-and-polymorphism/interfaces.rst
CarlyLanglois: actually now that you drew my attention here, some of this section and the stuff around it is wrong and not consistent with whats on the rest of the page. oops. i changed this bit just to encourage keeping the method public as a best practice

PR #34, opened 2020-04-06
Lsn9 exceptions content

CarlyLanglois:
MSBuild to come in a separate pr


src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,161 @@
+.. index:: ! exception, ! throw
+
+Exceptions
+==========
+
+Like most programming languages, C# includes exceptions and exception handling tooling. Some exception objects
+are provided by .NET itself, while others we may encounter from external C# libraries, or write ourselves.
```
gildedgardenia: This could be broken down to something like: "Some exception objects are provided by .NET itself. Other exception objects we may get from external C# libraries. We may even write these objects ourselves."

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,161 @@
...
+
+**Exceptions** in C# are objects derived from the ``System.Exception`` class. They are used in circumstances where something
+takes place outside of the ordinary flow of a program. When you want to explicitly call an exception in your code, you do so
```
gildedgardenia: What about "Exceptions occur at runtime when something takes place outside of the expected flow of our code."

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,161 @@
...
+
+**Exceptions** in C# are objects derived from the ``System.Exception`` class. They are used in circumstances where something
+takes place outside of the ordinary flow of a program. When you want to explicitly call an exception in your code, you do so 
+with the **throw** keyword. We've actually done this before in this book and will return to that example shortly.
+
+Maybe you see the term "exception" and instinctively groan or roll your eyes? Isn't programming only enjoyable when
+it works? Well, yes it is nice when your code performs as you expect and want it to. And in fact, exceptions are a 
+vital component of a healthy codebase. As a programmer, encountering an exception can provide crucial information 
+about proper usage of the. As a program writer, providing exceptions gives fellow programmers necessary intel on how your 
+app should run and be used.
```
gildedgardenia: This paragraph doesn't make a whole lot of sense to me as is. Perhaps, "While it would be nice if our code ran perfectly every time, exceptions are a key component of a codebase. When an exception arises, we, as programmers, can use the info to debug our code. For our fellow programmers, the exceptions we provide give them necessary intel on how our app runs and is meant to be used."

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,161 @@
...
+
+When Exceptions Arise
+---------------------
+
+You may have encountered exceptions in C# or another programming language already. Perhaps in on of these scenarios
+where exceptions often arise:
```
gildedgardenia: You could replace the final sentence with "... already, perhaps in one of the following scenarios: "

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,161 @@
...
+
+- Failure to connect to services external to your application, such as a database or API.
+- Failure to read or write to or from a file.
+- Failure to convert data, such as trying to convert something like ``"dog"`` to an ``int`` type. 
+- Null pointers. In other words, an object reference -- such as a method parameter or local variable -- that doesn't 
+  actually contain an object.
```
gildedgardenia: "Null pointers: where an object reference, like a variable, doesn't contain an object."

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,161 @@
...
+     
+	   
+Above, we see our Temperature constructor predictably sets the Fahrenheit value of ``insideTemp`` and 
+predictably throws an exception when provided a Fahrenheit value outside of the appropriate range. We don't 
+see any results of the print statement on the input's line 5 since the exception has caused the program 
+to stop running.  
+
+This is a common reason to include exception handling in your code. User input opens the door to a 
+variety of erroneous figures and good programs account for this uncertainty. Without exceptions in these 
+circumstances, a small typo could lead to any number of breaking errors down the stack trace.
```
gildedgardenia: I don't know if "breaking" is a necessary adjective. I think you could make the same point with just "errors"

gildedgardenia (commented):
I read through the content and I like the emphasis on why exceptions matter in particular. One thing I noted is that you write at a 10th grade level or higher. My concern is that some students may miss a point if they do not have the reading comprehension skills to read at this level. Would you take a second look at this content before I re-review?

src/chapters/exceptions/introduction.rst
CarlyLanglois: yes this is great thanks. i see a major typo in the original text too. no wonder it didnt make sense 😆

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,169 @@
...
+
+Exceptions
+==========
+
+Like most programming languages, C# includes exceptions and exception handling tooling. 
+Some exception objects are provided by .NET itself. Other exception objects we may get from 
+external C# libraries. We may even write these objects ourselves.
+
+**Exceptions** in C# are objects derived from the ``System.Exception`` class. 
+Exceptions occur at runtime when some event takes place outside of the expected flow of our code.
```
gildedgardenia: I went over this phrasing a couple of times and I think these two sentences seem slightly at odds with each other. The first one states that exceptions are objects and the second one implies that exceptions are events. Maybe rephrase the second sentence to "Calls to exceptions occur...."

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,169 @@
...
+
+- Failure to connect to services external to your application, such as a database or API.
+- Failure to read or write to or from a file.
+- Failure to convert data, such as trying to convert something like ``"dog"`` to an ``int`` type. 
+- Failure to reach an actual object. 
+
+	.. admonition:: Note 
+	
+		This last scenario is a **null pointer**. This is an object reference, 
+		like a variable, that doesn't actually contain an object.
```
gildedgardenia: I don't think you need "like a variable" here. The sentence is just as effective reading "A null pointer is an object reference that doesn't actually contain an object."

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,169 @@
...
+
+This is a common reason to include exception handling in your code. User input opens the door to a 
+variety of erroneous figures and good programs account for this uncertainty. Without exceptions in these 
+circumstances, a small typo could lead to any number of errors down the stack trace. 
+
+.. index:: ! exception handling
+
+When we throw an exception like in the example above, we flag the anomalous circumstance. If we choose to 
+do nothing when the exception is thrown, the program will stop and a record of the exception
+can be found in the stack trace. Alternatively, we can **handle** an exception and offer an alternative
```
gildedgardenia, line 119: This sentence order is a little confusing. Maybe change "If we had chosen to do nothing.... stack trace. Instead, we handled..."

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,169 @@
...
+      at lsn4_demos.Temperature..ctor(Double fahrenheit) in /Users/carlylanglois/launchcode/csharp/web-dev-exercises/lsn4-demos/lsn4-demos/Temperature.cs:line 40
+      at lsn4_demos.Program.Main(String[] args) in /Users/carlylanglois/launchcode/csharp/web-dev-exercises/lsn4-demos/lsn4-demos/Program.cs:line 14
+     
+	   
+Above, the Temperature constructor predictably sets the Fahrenheit value of ``insideTemp`` and 
+throws an exception when provided a Fahrenheit value outside of the appropriate range. We don't 
+see any results of the print statement on the input's line 5 since the exception has caused the program 
+to stop running.  
+
+This is a common reason to include exception handling in your code. User input opens the door to a
```
gildedgardenia: So I reread this paragraph after the paragraph below and now the order really does not make sense. I would switch the paragraph below to above these two paragraphs. That order would read: "We can do this thing called exception handling" -> "Above the program stopped running" -> "That is why we use exception handling"

src/chapters/exceptions/introduction.rst
```diff
@@ -0,0 +1,169 @@
...
+action, bypassing the need to stop the program. We'll cover how to handle exceptions on the next page.
+
+When to Use Exceptions
+----------------------
+
+It is wise to use an exception if you find that there is some level of chance involved in your 
+program. This could be a situation where a variable is dependent on user input or a connection to 
+another service.
+
+You may want to address the uncertainties in a different fashion. With our temperature app for example, rather than
```
gildedgardenia: "the uncertainties" is very vague. Is this addressing the uncertainties of working with user input?

src/chapters/exceptions/handling.rst
```diff
@@ -0,0 +1,189 @@
+.. index:: ! checked exception, ! unchecked exception
+
+Handling Exceptions
+===================
+
+When you include exceptions in your C# programs, you must decide what should take place once one is thrown. Some languages, Java 
+for instance, require exceptions to be handled at the time of compiling. These are **checked exceptions**. When an exception is
```
gildedgardenia: "Some languages, like Java, require exceptions..."

src/chapters/exceptions/handling.rst
```diff
@@ -0,0 +1,189 @@
+.. index:: ! checked exception, ! unchecked exception
+
+Handling Exceptions
+===================
+
+When you include exceptions in your C# programs, you must decide what should take place once one is thrown. Some languages, Java 
+for instance, require exceptions to be handled at the time of compiling. These are **checked exceptions**. When an exception is 
+not handled in compiling, it is passed to runtime and called an **unchecked exception**. C# does not contain checked exceptions
```
gildedgardenia: Maybe clarifying "C# does not contain either checked or unchecked exceptions... " or "Exceptions in C# are all unchecked exceptions. Therefore, it is up to you, ..."

src/chapters/exceptions/handling.rst
```diff
@@ -0,0 +1,189 @@
...
+
+Here's some advice to consider when contemplating when to handle an exception. This comes from Karl Seguin's 
+`Foundations of Programming <https://www.openmymind.net/FoundationsOfProgramming.pdf>`__.
+
+#. Only handle exceptions that you can actually do something about.
+#. You can't do anything about the vast majority of exceptions.
+
+For example, if your code cannot connect to a database, there is probably nothing your program can 
+do about it. However, as we allude to on the previous page, if you receive invalid input from a user, you can still throw an 
+exception, re-prompt them to refine their input with an error message to help them get it right the next time.
```
gildedgardenia: "... exception and re-prompt them to refine ..."

src/chapters/exceptions/handling.rst
```diff
@@ -0,0 +1,189 @@
...
+      {
+         Fahrenheit = fahrenheit;
+      }
+      catch(ArgumentOutOfRangeException e)
+      {
+         Fahrenheit = -459.67;
+      }
+   }
+
+The first action the constructor method does is call the Fahrenheit's setter method. If invoking the
```
gildedgardenia: "call ``Fahrenheit``'s setter method."

src/chapters/exceptions/handling.rst
```diff
@@ -0,0 +1,189 @@
...
+
+The first action the constructor method does is call the Fahrenheit's setter method. If invoking the 
+setter does not result in an exception being thrown, then the new ``Temperature`` object is created and
+given an initial Fahrenheit value. If the action inside the ``try`` block results in an exception,
+or specifically an ``ArgumentOutOfRange`` instance, then the ``catch`` block runs and the initial 
+value of a new ``Temperature`` object's ``fahrenheit`` field is set to absolute zero. 
+
+Although in this example, we don't explicitly do anything with the argument ``e`` ("e" for exception),
+there are some scenarios where you may wish to display the exception or it's message in the view of your
+running app. It also an option to rethrow an exception after it has been caught. You won't need to rethrow
```
gildedgardenia: This sentence is really long. Perhaps break it up like so: "With this example, we don't explicitly do anything with the argument ``e`` ("e" for exception). However, there are scenarios ...."

src/chapters/exceptions/common-exceptions.rst
```diff
@@ -0,0 +1,84 @@
+Common Exception Objects
+========================
+
+Below is a summary of some of the more commonly used exception types in C#. As we mention before, all exceptions extend the 
+``System.Exception`` class. It is also possible to write your own exception type that inherits from ``System.Exception``. You 
+may find that your particular cause of error elicits a custom exception type.
```
gildedgardenia: I would move these last two sentences into your admonition right below them.

gildedgardenia (commented):
I really like the improvements you made to the chapter! As we discussed earlier, I do think having the Temperature program in their repo will be helpful. Other than that, my comments are mainly around wording that I found confusing.

src/chapters/exceptions/introduction.rst
CarlyLanglois: im assuming this comment is out of date with your understanding now?

src/chapters/exceptions/introduction.rst
CarlyLanglois: yes and the other example in above. just changed it to "those uncertainties" to keep clear that we are talking about the items just mentioned

src/chapters/exceptions/handling.rst
CarlyLanglois: good note!

gildedgardenia (approved):
Looks great!!

PR #37, opened 2020-04-22
Lsn10 studio

CarlyLanglois:


src/chapters/razor-views/studio.rst
```diff
@@ -0,0 +1,120 @@
...
+
+#. Iteratively add the values in ``appropriateFacials`` to an unordered list. If you need a quick reminder of
+   the syntax, review the :ref:`@foreach section <razor-foreach>` .
+
+Mani or Pedi?
+-------------
+
+We want to display a description for the nail service the user selects.
+Inside the ``servicesMenu`` div, use ``@if`` to determine if the value of 
+``manipedi`` is a ``"manicure"`` or ``"pedicure"``.
```
gildedgardenia: The "a" is unnecessary here.

src/chapters/razor-views/studio.rst
```diff
@@ -0,0 +1,120 @@
...
+   :alt: Profile showing Yolanda's information, recommended facial treatments, and pedicure description.
+
+   The client's service menu.
+
+Bonus Mission
+-------------
+
+#. Modify ``site.css`` to get some CSS practice. Try add a footer with square
+   shaped ``div`` elements. Each square should be a different color for
+   different available nail polishes.
```
gildedgardenia: Since ``site.css`` has a lot of CSS that Microsoft in it, I might add a suggestion as to where to put this. I think some students would like this challenge but might be deterred from trying it because of how much is already going on in ``site.css``.

gildedgardenia (approved):
Really love this studio! I left some suggestions, but whenever you think this studio is ready, merge away!

PR #38, opened 2020-04-23
Lesson 9 Exercises

gildedgardenia:


src/chapters/exceptions/exercises.rst
```diff
@@ -0,0 +1,29 @@
+Exercises
```
CarlyLanglois: This should be Exercises: <section topic>. check the other exercise headers for the format

src/chapters/exceptions/exercises.rst
```diff
@@ -0,0 +1,29 @@
...
+=========
+
+To get started, fork and clone this `repo <https://github.com/LaunchCodeEducation/csharp-web-dev-lsn9exceptions>`_ from Github.
```
CarlyLanglois, line 4: Can you make this repo public and not private please?

src/chapters/exceptions/exercises.rst
```diff
@@ -0,0 +1,29 @@
+Exercises
+=========
+
+To get started, fork and clone this `repo <https://github.com/LaunchCodeEducation/csharp-web-dev-lsn9exceptions>`_ from Github. 
+
+Divide By Zero!
+---------------
+
+Complete a function called ``Divide()`` in ``Program.cs``.
```
CarlyLanglois: image
the way these two sentences get rendered together makes Program.cs.Divide() look like one entity. Change the second to say "the method..."?

src/chapters/exceptions/exercises.rst
```diff
@@ -0,0 +1,29 @@
...
+``Divide()`` takes in two parameters: ``x`` and ``y``.
+
+Your function should return the result of ``x/y``.
+
+However, if ``y`` is zero you should throw an appropriate error message.
```
CarlyLanglois: change error message to exception. The exception will include a custom error message that they write.

src/chapters/exceptions/exercises.rst
```diff
@@ -0,0 +1,29 @@
...
+Try to use an ``ArgumentOutOfRangeException`` and put your ``try/catch`` block in ``Main()`` to test out your error-handling skills!
+
+Test Student Labs
+-----------------
+
+After mentioning to Professor Jackson that you would like to get some more practice with exceptions, she offered to let you write some grading software!
```
CarlyLanglois, line 25: who the heck is this? can you provide more context here if you're telling a story. is this meant to be a theme through the whole page or specifically this task?

in the same vein, is there a connection between these two methods? if there is none, consider placing these in separate files and wrapper classes. we've spent the last several lessons introducing object oriented concepts and it makes sense to practice what we preach.

src/chapters/exceptions/exercises.rst
```diff
@@ -0,0 +1,29 @@
...
+Divide By Zero!
+---------------
+
+Complete a function called ``Divide()`` in ``Program.cs``.
+``Divide()`` takes in two parameters: ``x`` and ``y``.
+
+Your function should return the result of ``x/y``.
+
+However, if ``y`` is zero you should throw an appropriate error message.
+Try to use an ``ArgumentOutOfRangeException`` and put your ``try/catch`` block in ``Main()`` to test out your error-handling skills!
```
CarlyLanglois: This goes for both of these methods - I think it's a good idea to say explicitly that if an exception is caught, then they should print out the error message

CarlyLanglois (approved):
Good. I like these two problems! I do think you can tie them together if you want to extend the story that begins in the second part, or make them more distinct with classes - or a combination of both.
Also, just putting this on the table, there are a lot exclamation points on this page.

src/chapters/exceptions/exercises.rst
gildedgardenia: Fixed!

src/chapters/exceptions/exercises.rst
gildedgardenia: Fixed!

src/chapters/exceptions/exercises.rst
gildedgardenia: Done!

src/chapters/exceptions/exercises.rst
gildedgardenia: Done!

src/chapters/exceptions/exercises.rst
gildedgardenia: Changed intro and now the two are better connected!

src/chapters/exceptions/exercises.rst
gildedgardenia: Done!

PR #39, opened 2020-04-24
Lesson 9 Studio

gildedgardenia:


src/chapters/vstools/studio.rst
```diff
@@ -0,0 +1,22 @@
+If It Ain't Broke, Add a Breakpoint!
+====================================
+
+On your machine, open up your fork of the ``HelloASPDotNETDemo`` `repository <https://github.com/LaunchCodeEducation/HelloASPDotNETDemo>`_ in Visual Studio.
+The purpose of this studio is to talk about your current debugging strategies and how to make the most of the debugger tools discussed in this chapter.
+
+Before we start practicing with debugging tools, go over with the group one error you encountered when working on your own version of ``HelloASPDotNET``.
```
CarlyLanglois, line 12: how about adding some suggestions of types of errors they may have seen to get the ball rolling?

src/chapters/vstools/studio.rst
```diff
@@ -0,0 +1,22 @@
...
+
+#. What was the error?
+#. How did you solve this error? What have been the strategies and tools you have been using so far to debug your code?
+#. Could one of the debugging tools helped you when addressing this error?
```
CarlyLanglois: help instead of helped?

src/chapters/vstools/studio.rst
```diff
@@ -0,0 +1,22 @@
...
+
+Review the code from ``HelloASPDotNETDemo`` and use the debugging tools in Visual Studio to address potentional problems.
+
+Questions to review as you look through the code:
+
+#. When would you use the *Call Stack* pane? If you run the app and it is already functioning, what shows up there? 
+#. What breakpoints should you add?
+#. Would a conditional breakpoint make sense to use in the context of this app?
+
+Once you have gone through the demo code, open up a piece of code you have been struggling with.
```
CarlyLanglois: this is a good prompt to get them to think about other programs that they might be struggling with

src/chapters/vstools/studio.rst
```diff
@@ -0,0 +1,22 @@
...
+On your machine, open up your fork of the ``HelloASPDotNETDemo`` `repository <https://github.com/LaunchCodeEducation/HelloASPDotNETDemo>`_ in Visual Studio.
+The purpose of this studio is to talk about your current debugging strategies and how to make the most of the debugger tools discussed in this chapter.
+
+Before we start practicing with debugging tools, go over with the group one error you encountered when working on your own version of ``HelloASPDotNET``.
+
+#. What was the error?
+#. How did you solve this error? What have been the strategies and tools you have been using so far to debug your code?
+#. Could one of the debugging tools helped you when addressing this error?
+
+Review the code from ``HelloASPDotNETDemo`` and use the debugging tools in Visual Studio to address potentional problems.
```
CarlyLanglois: For this section, im having trouble envisioning many students actually doing this if we don't provide specific items to look for. I think we talked about like suggesting lines to add breakpoints and looking for specific variables? adding a certain watch someplace maybe?

src/chapters/vstools/studio.rst
```diff
@@ -0,0 +1,22 @@
+If It Ain't Broke, Add a Breakpoint!
+====================================
+
+On your machine, open up your fork of the ``HelloASPDotNETDemo`` `repository <https://github.com/LaunchCodeEducation/HelloASPDotNETDemo>`_ in Visual Studio.
```
CarlyLanglois: can you clarify, why cant they do this studio with their own app and not the demo version?

src/chapters/vstools/studio.rst
gildedgardenia: Added some things that could have gone for students with the application!

src/chapters/vstools/studio.rst
gildedgardenia: Ah very true! Thank you!

src/chapters/vstools/studio.rst
gildedgardenia: Thank you! I am hoping that it will divert students who want to work on other stuff back to thinking about debugging by giving them the space to work on other stuff.

src/chapters/vstools/studio.rst
gildedgardenia: That escaped my mind! I went back and added a direction to get them started and an example to one of the questions so that discussion can flow more easily! Thank you!

src/chapters/vstools/studio.rst
gildedgardenia: I got all messed up! This is a mistake! I was supposed to put the demo code in a note that specified in students were struggling with getting their own version of the app to work, they could use the demo code for the purposes of the first part of the studio.

CarlyLanglois (approved):
Can you change the page title to match the other studios? Studio: Studio Name

Rather than this being a thought-inducing exercise for the students, I had pictured from your initial description that this studio would include specific tasks for the students to do. Similar to how you ask to add *name* to the watch pane. You could expand on this idea asking them specific questions about, if they provide a value of "sally", when/what line in what mvc file is this assigned? other variables?
If a breakpoint is added to line x, what do expect to be the value of that variable? what would the last view of the running app be before it stops running? 
what if a breakpoint is added to line y? z?
when you add that breakpoint on line whatever, what is the last line in your call stack?
if you want to look for how the value of variable v is assigned, what line would add a breakpoint?
think of a scenario where a conditional breakpoint could be handy here and tell them to do it and look for a particular outcome.
I think that our lift needs to be higher here because there is something inherently dry and not inviting about debugging tools. would you be voluntarily looking for places to add breakpoints yourself?
If we can get them engaged though, they will benefit from a bit of practice and exposure

PR #42, opened 2020-05-05
add model intro page

CarlyLanglois:
pausing the rest of this section for now but this page is virtually unchanged from Java and ready to be reviewed now


src/chapters/aspdotnet-model-classes/model-intro.rst
```diff
@@ -0,0 +1,124 @@
...
+the views. This data needs to come from some source and take some shape. Cue the models.
+
+What is a Model?
+-----------------
+
+.. index:: model
+
+A **model** represents the logic for accessing and storing the data used in an application. 
+Properly constructed, they do not depend on any controllers or views, which
+makes models easy to reuse without modification.
```
gildedgardenia: The use of "they" as the subject makes this sentence a little ambiguous. I would switch it around so it goes something like, "..., models do not depend ..., which makes them easy to ..."

src/chapters/aspdotnet-model-classes/model-intro.rst
```diff
@@ -0,0 +1,124 @@
...
+What is a Model?
+-----------------
+
+.. index:: model
+
+A **model** represents the logic for accessing and storing the data used in an application. 
+Properly constructed, they do not depend on any controllers or views, which
+makes models easy to reuse without modification. 
+
+Models are not the data itself, but rather the data logic. They dictate how we want to handle
```
gildedgardenia: I think this might make more sense as "Models do not contain data, but rather the data logic." Since the rest of the paragraph goes on to describe how data is stored in something like a database and then that info is passed up to a model for shaping. I also am not sure about the phrase "data logic". If I was a student and tried to google that to get some clarification, then the first result would pertain to a company called Datalogic. Maybe use this sentence as well to briefly touch on what students should be thinking of when they see data logic.

src/chapters/aspdotnet-model-classes/model-intro.rst
```diff
@@ -0,0 +1,124 @@
...
+or an external data service. Data is typically application-agnostic and it is the work of the models
+we write to shape raw data into useful application information.
+
+Consider a physical address book (a model). It stores names, addresses, phone
+numbers, etc. on its pages. Anyone (a controller) can pick up the book,
+retrieve the information, and then write some of the data on a separate piece
+of paper (a view). The address book model does not depend on who picks it up and enters
+their contacts. The book just provides organization and storage logic. On the flip side, the same
+person can input the same contact data into a different book. So a model takes raw information and 
+turns it into something useful for a particular application.
```
gildedgardenia: I like the example, but the way it is currently written makes it seem in direct contradiction with what is written above about how models do not store data. Maybe instead of phrasing it as "It stores names, addresses, ... ", you could instead phrase it as "The pages are laid out so each section contains a blank for the name, address, phone number, and other important information we need to contact a friend."

src/chapters/aspdotnet-model-classes/model-intro.rst
```diff
@@ -0,0 +1,124 @@
...
+
+#. "Please retrieve this information from the model.",
+#. "Please store this according to the rules of the model.",
+#. "Please delete this item.",
+#. "Please change this item.",
+#. "Please display this.",
+#. "Please modify this data in a novel way".
+
+Remember, the controllers are the traffic cops, directing information from one place to another. 
+A controller does not permanently store data. Instead, it either
```
gildedgardenia: This sentence makes it seem like a model does store data, which this page emphasizes that a model does not store the data. Maybe rephrase to "A controller does not contain data logic"

gildedgardenia (commented):
I like the way the page is laid out and really drives home the key points about what makes a model different from the other components in an MVC application! However, I did find the page to be written at a high reading level. One of the paragraphs I tested was rated a grade 10 reading level.

CarlyLanglois:
@gildedgardenia this is ready for another re-review if you need it. i made some hemingway suggested changes but I think aiming for it's 5th grade reading level on some of this material could be a challenge

src/chapters/aspdotnet-model-classes/model-intro.rst
```diff
@@ -0,0 +1,122 @@
+Models in MVC
+==============
+
+In the previous chapter, you learned about Razor, which displays data and an
+interface for a user. Before that, you learned about :ref:`controllers`. Remember that 
+determine what data to send to the the views. This data needs to come from some source
```
gildedgardenia: Is this supposed to read "Remember that controllers determine what data is sent to the views."?

src/chapters/aspdotnet-model-classes/model-intro.rst
```diff
@@ -0,0 +1,122 @@
...
+
+   Need further review? Check out `MVC for Noobs <https://code.tutsplus.com/tutorials/mvc-for-noobs--net-10488>`__.
+
+Model vs. Controller
+--------------------
+
+At this point, it might be tough to decide what code belongs in the model and what belongs in the 
+controller. Here are some general guidelines. Any code that transfers data or responds to user 
+actions belongs in the controller. Code that retrieves data from a database or organizes that 
+data belongs with the model.
```
gildedgardenia, line 67: You can eliminate "Here are some general guidelines." and start the following sentence with "In general, any ..."

gildedgardenia (approved):
I really like this page! I think it is really clear and concise!

src/chapters/aspdotnet-model-classes/model-intro.rst
CarlyLanglois: this change is from the hemingway suggestions 😝

PR #44, opened 2020-05-08
Lsn10 create template

CarlyLanglois:
2nd page of views lesson


src/chapters/razor-views/create-template.rst
```diff
@@ -0,0 +1,128 @@
...
+Razor
+-----
+
+.. index:: ! Razor
+
+**Razor** is a templating syntax included in the application framework for our MVC project. 
+It allows us to write C# code directly into an HTML tree. 
+
+More information on Razor can be found 
+`here <https://docs.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-3.1>`__.
```
gildedgardenia: "here" is not a great way for a link because it doesn't indicate much about what the student is about to click on. I would recommend changing it to "documentation", "article", "resource", etc. Then just reword the sentence a bit so the new word makes sense!

src/chapters/razor-views/create-template.rst
```diff
@@ -0,0 +1,128 @@
...
+
+For the most part, Razor templates look and operate just like
+regular HTML code. Any C# logic we add to a template is preceded with 
+the ampersand symbol, ``@``. You can open any template in a browser and view it just
```
gildedgardenia: A couple notes on this sentence:

- The at sign doesn't have a commonly used word so I would just call it "the at sign". I think if you really want to use something else, asperand is a good one. But that being said, with the at sign being so commonly referred to as such, more students will understand it that way. The ampersand symbol is &.
- I read this and immediately wanted to see an example. Maybe direct students to when they will get to see this syntax? Some students may need that aid to understand what is going on here.

src/chapters/razor-views/create-template.rst
```diff
@@ -0,0 +1,128 @@
...
+   video is found on the `views-static branch <https://github.com/LaunchCodeEducation/HelloASPDotNETDemo/tree/views-static>`__.
+
+The sections below outline the main ideas presented in the video. However, the
+text is NOT a substitute for completing the work described in the clip.
+
+.. index:: ! View() method
+
+Controllers Return View Templates
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
gildedgardenia, line 57: I think this is a missed opportunity to draw a connection between what we said about ``IActionResult`` in Lesson 8. We briefly mentioned that they could return an HTML template with that return type in an action method.

src/chapters/razor-views/create-template.rst
```diff
@@ -0,0 +1,128 @@
...
+**WINDOWS USERS:** Right click on the ``Views`` directory or any subdirectory and select *Add -> New Item*. 
+Avoid selecting *Add -> View* from your options. 
+
+**MAC USERS:** Right click on the ``Views`` directory or any subdirectory and select *Add -> New File*.
+
+**ALL USERS:** In the modal window that appears, look for the item labelled *Razor View*. 
+
+.. figure:: figures/razor-view-template-selection.png
+   :alt: User selects the *Razor View* file template.
```
gildedgardenia: I would remove these asterisks. A screen reader is going to read this as "asterisk Razor" and it might sound confusing.

src/chapters/razor-views/create-template.rst
```diff
@@ -0,0 +1,128 @@
...
+YOUTUBE VIDEO HERE
+
+.. admonition:: Note
+
+   The starter code for this video is found at the `attribute-routing branch <https://github.com/LaunchCodeEducation/HelloASPDotNETDemo/tree/attribute-routing>`__
+   of ``HelloASPDotNETDemo``. The final code presented in this 
+   video is found on the `views-static branch <https://github.com/LaunchCodeEducation/HelloASPDotNETDemo/tree/views-static>`__.
+
+The sections below outline the main ideas presented in the video. However, the
+text is NOT a substitute for completing the work described in the clip.
```
gildedgardenia: This paragraph combined with the warning above feels like too much. Maybe just one warning?

src/chapters/razor-views/create-template.rst
```diff
@@ -0,0 +1,128 @@
...
+**ALL USERS:** In the modal window that appears, look for the item labelled *Razor View*. 
+
+.. figure:: figures/razor-view-template-selection.png
+   :alt: User selects the *Razor View* file template.
+
+   Select the *Razor View* file template.
+
+
+Add the form HTML to the template.
+Update the ``HelloController`` method ``Index()`` to return ``View()``.
```
gildedgardenia: I would expand on these sentences. "Add the HTML form from the ``Index()`` method to your new template. Once you are done with that, update the ``Index()`` method so that it only returns ``View()``." When I first read this section, I missed what was going on here. Some learners who really count on the written text to drive home their understanding may also miss it.

src/chapters/razor-views/create-template.rst
```diff
@@ -0,0 +1,128 @@
...
+
+Add the form HTML to the template.
+Update the ``HelloController`` method ``Index()`` to return ``View()``. 
+
+When you re-render the app now, you'll notice some additional features and different styling.
+There is a ``_Layout.cshtml`` file inside of the ``Views/Shared`` folder that provides some scaffolding 
+for the application views. You can edit this file once and any other template will also render whatever 
+elements you add to it. This can help give your application a polished, unified look and feel.
+If you do not want to use this file, adding ``@{Layout = null;}`` at 
+the top of the template will cause this shared layout template to be ignored.
```
gildedgardenia, line 114: I am not sure about this paragraph. I like where you are going with it and I think it drives home an important point! They just added the form to the template so now when they load the form page it is using the template which in turn uses ``_Layout.cshtml``. However, it took me a second to understand this point. My misunderstanding could be due to the fact that the app isn't running in front of me as I write this so I don't have the visual aide that the students will. However, from the student perspective, if I came back to this to simply review something from a later lesson, I would feel a little lost.

src/chapters/razor-views/create-template.rst
```diff
@@ -0,0 +1,128 @@
...
+``HomeController`` action methods. Just as an action method's name will map to a route with the same 
+name, an action method's name can also correspond to a Razor template's name. The ``Index()`` method 
+inside of ``HomeController`` returns the ``Index.cshtml`` template found inside of ``Views/Home``.
+
+In order to return that template, the action method calls a **View() method**. The ``View()`` method finds 
+a template that's associated with the particular controller and action method that it is called from. 
+
+To override the default behavior of the ``View()`` method, you can pass in a parameter which is the name 
+of the template you want to render. For example, if we want an action method named ``Form()`` to return a 
+Razor template named ``WelcomeForm.cshtml``, we have ``Form()`` return ``View("WelcomeForm");``.
```
gildedgardenia: I might move this into a note admonition. I think it is important knowledge, but I am not sure how much students will be using it in this course.

gildedgardenia (commented):
This is looking great so far! I like the outline for the video. I left some grammar-related comments, but most of the comments are just my opinion and experience reading the text.

src/chapters/razor-views/create-template.rst
CarlyLanglois: yeah actually i feel the same way - i almost just left this in a comment to figure out when the rest of the chapter is fleshed out. 
I'll add a todo to return to this question later but keep the short intro here for now

gildedgardenia: I think keeping this in will work! I reviewed the first video I made and the second one and I think my temptation after showing how you can "turn off" Layout to then bring it back was really why it felt so jumpy. I also like the intro to the idea of what is going on the ``Shared`` directory and why the files in there are important. I feel confident keeping the intro you have written here and adding that tidbit in the vid.

gildedgardenia (approved):
Looks great! If you want to merge, I can create a PR later with video links and any additional text needed for what is going on in the video.

PR #45, opened 2020-05-08
Lsn10 using templates

CarlyLanglois:
third page - this one is kind of short using the java format and depending on how the next couple pages pan out, I think it could be combined into other content.

I also opted not to address ViewData right now. Its seems like we can get away with just using viewbag in this lesson and/or just mention that viewdata acts similarly. once we introduce models we'll want to move away from both of these


src/chapters/razor-views/using-templates.rst
```diff
@@ -0,0 +1,90 @@
+Using a Template
+================
+
+This page and the ones that follow introduce how to start to pass data between the MVC elements.
+We haven't addressed models yet (those come next) so this just means sharing information between 
+the controller and view.
```
gildedgardenia: I would rephrase this. Maybe something like, "Now that we know a little bit about views, we can start talking about how to pass data between MVC elements. Models are a key component of this, but for now, we will focus on how to pass data between the view and the controller."

src/chapters/razor-views/using-templates.rst
```diff
@@ -0,0 +1,90 @@
...
+   {
+      // method code here
+      
+      ViewBag.property = Data;
+      return View();
+   }
+
+**ViewBag** is an object that passes data into a template. ``Data`` can
+be a variable of any type, a number, a collection of some sort, or an object. 
+A ``ViewBag`` property is created and given a value as simply as is done on line 5 above.
```
gildedgardenia: I think this explanation could be expanded. Right now, it reads like ViewBag is a set system object with set properties and if I found the right page in the docs I could see the list of available properties. This [article](https://www.tutorialsteacher.com/mvc/viewbag-in-asp.net-mvc) on ViewBag had a few points that really helped me figure out what ViewBag is, such as it is dynamic and it is used to pass data from the controller into the view. Also, it is a dynamic dictionary so knowing that helped me realize that I could name the properties whatever I wanted to because the properties act as the keys.

src/chapters/razor-views/using-templates.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Accessing Data in a Template
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The data assigned to properties on ``ViewBag`` is available inside of Razor templates.
+It can be accessed with the syntax ``@ViewBag.property``.
+
+For example, if the controller stores data in ``ViewBag.name``, then the
+template can display that value like so:
```
gildedgardenia: It feels like there is a bit of disconnect between this sentence and the preceding one. Maybe expand to "For example, we have a property called ``name`` and the controller is storing some string in ``ViewBag.name``. We could pass that string to the template using the at sign: "

gildedgardenia (commented):
Overall, I think this page is great! I like the concept check questions and I like the way the page drives home that if you change a value, the view reflects that change. I do think that adding some additional explanation around what ViewBag is would be helpful to students.

src/chapters/razor-views/using-templates.rst
CarlyLanglois: I changed some wording around this example and used a different property name. interested to know if the change I made from the comment just before this - on how to create and assign property values - will help to clarify this section also. ``name`` itself is nothing before being created as a property of ``ViewBag``

gildedgardenia (approved):
This is awesome! Thanks for adding in the extra clarification!

src/chapters/razor-views/using-templates.rst
gildedgardenia: I think the new wording is great! It really helps clarify the connection and it reinforces to me as the reader that I create the properties and I have some sense of ownership over what the properties of ``ViewBag`` are.

PR #46, opened 2020-05-11
Lsn10 template iteration

CarlyLanglois:


src/chapters/razor-views/template-iteration.rst
```diff
@@ -0,0 +1,247 @@
...
+.. index:: ! var
+
+Nested Loops
+------------
+
+Assume you have a collection of different ``CoffeeShop`` objects. Each 
+object contains a string field for ``name`` and a field that is a list of 
+of the brews available, ``coffeeOptions``. 
+
+Below, we use nest loops to display a list of the shop names and their brew
```
gildedgardenia: "nested"

src/chapters/razor-views/template-iteration.rst
```diff
@@ -0,0 +1,247 @@
...
+   </ul>
+
+   <p>Brews Brothers</p>
+   <ul>
+      <li>French Roast</li>
+      <li>Kopi Luwak</li>
+   </ul>
+
+
+Apart from the nest loops displayed above, here are ome other items you may find useful
```
gildedgardenia: "some"

src/chapters/razor-views/template-iteration.rst
```diff
@@ -0,0 +1,247 @@
...
+  code block above. Comments in Razor are nested between ``@*`` and ``*@``.
+  You may have noticed the comment block present on the top of a new view file.
+
+.. TODO: is this a sufficient note for the use of var here? It feel unfinished but im
+at a loss for what else is needed here.
+
+- ``ViewBag.coffeeShops`` is a list of ``CoffeeShop`` objects but we've used ``var``
+  on line 1 to type the ``coffeeShop`` item. In some limited circumstances, we can 
+  use the **var** keyword to implicitly type a variable.
+  Alternatively, Razor does also allow us to import a custom class. such as ``CoffeeShop``.
```
gildedgardenia: "class, such as"

src/chapters/razor-views/template-iteration.rst
```diff
@@ -0,0 +1,247 @@
...
+
+Apart from the nest loops displayed above, here are ome other items you may find useful
+to note from the example above. 
+
+- Razor comments are seen on lines 3 and 8 in the first 
+  code block above. Comments in Razor are nested between ``@*`` and ``*@``.
+  You may have noticed the comment block present on the top of a new view file.
+
+.. TODO: is this a sufficient note for the use of var here? It feel unfinished but im
+at a loss for what else is needed here.
```
gildedgardenia: I like the definition. Maybe acknowledge that this is unusual for a statically typed system? I think you could also add a reason why you used ``var``. Was it to avoid having to import the ``CoffeeShop`` class? Why is that advantageous?

src/chapters/razor-views/template-iteration.rst
```diff
@@ -0,0 +1,247 @@
...
+
+- ``ViewBag.coffeeShops`` is a list of ``CoffeeShop`` objects but we've used ``var``
+  on line 1 to type the ``coffeeShop`` item. In some limited circumstances, we can 
+  use the **var** keyword to implicitly type a variable.
+  Alternatively, Razor does also allow us to import a custom class. such as ``CoffeeShop``.
+  Another way to use this list of ``CoffeeShop`` objects would be to import the class
+  at the top of the template with a :ref:`using-statement` statement.
```
gildedgardenia: After re-reading the paragraph, I would rephrase this to something like: "Alternatively, Razor does allow us to import a custom class, such as ``CoffeeShop``. If we wanted to do so, we could import the class at the top of the template ..."

src/chapters/razor-views/template-iteration.rst
```diff
@@ -0,0 +1,247 @@
...
+               </ol>
+            </div>
+         }
+
+   Which option produces:
+
+   #. One heading, 4 ordered lists, and 4 coffee names (each name labeled as "1")?
+   #. One heading, one ordered list, and 4 coffee names (with the names labeled
+      1, 2, 3...)?
+   #. 4 headings, 4 ordered lists, and 4 coffee names (each name labeled as "1")?
```
gildedgardenia: After reading the whole page, this feels disjointed. The questions serve better as concept checks. I would shorten this warning to just reflect that location matters and give one example of something looking off because a dev chose the wrong locations in their template. Then move the code blocks next to the questions in the concept checks. That might also make it easier to import into Canvas.

gildedgardenia (commented):
Awesome examples! I add some suggestions for the TODO you left. I hope the suggestions are helpful!

src/chapters/razor-views/template-iteration.rst
CarlyLanglois: im still not sure if there is an advantage tbh. just added more info to the definition and more context to that todo comment so we can return to it when needed

gildedgardenia (approved):
Thanks for separating out those examples! Looks great!

PR #47, opened 2020-05-12
Lsn10 template conditionals

CarlyLanglois:


src/chapters/razor-views/template-conditionals.rst
```diff
@@ -0,0 +1,171 @@
...
+
+Above, ``condition`` represents any expression that can be evaluated to ``true`` or ``false`` 
+(ie, a boolean).
+
+If ``condition`` evaluates to ``true``, then Razor adds the HTML element to
+the webpage and the content is displayed in the view. If ``condition`` is
+``false``, then Razor does NOT generate the element, and the content stays
+off the page.
+
+.. admonition:: Example
```
gildedgardenia: This kind of jumped for me. The paragraph ends with the idea that content doesn't end up on the page and then transitions to an example where ``else`` is used to make sure alternate content ends up on the page. Maybe if it transitions with some sentence that notes that we can also use a larger if/else construction to control which content appears on the page, it wouldn't read as disconnected?

src/chapters/razor-views/template-conditionals.rst
```diff
@@ -0,0 +1,171 @@
...
+view. If they do not match, then the condition on line 5 is evaluated. If that 
+``else if`` statement is true, then a paragraph element with different internal
+text is added. Lastly, if neither of those conditions is met, then the paragraph
+element with text on line 11 is rendered. 
+
+.. admonition:: Note
+
+   We don't add an additional ``@`` symbols on lines 5 and 9 in the 
+   example above. ``if``, ``else if``, and ``else`` clauses are dependent
+   and treated as one C# code block in Razor.
```
gildedgardenia, line 67: I love this note! I always mess this up and add too many at signs and end up with build errors.

gildedgardenia (approved):
I think this page looks great! Whenever you feel good about it, please merge away! I can go back and add the video and any supporting text later.

PR #48, opened 2020-05-14
Lsn11 model class page

CarlyLanglois:
2nd intro to models page


src/chapters/aspdotnet-model-classes/model-class.rst
```diff
@@ -0,0 +1,224 @@
+Create a Model
+==============
+
+In the next several pages, we will be making updates to ``CodingEvents`` to demonstrate model creation,
```
gildedgardenia: I think "over" makes more sense than "in" here.

src/chapters/aspdotnet-model-classes/model-class.rst
```diff
@@ -0,0 +1,224 @@
...
+how models relate to data, and the practice of model binding. The first of these steps is to move data 
+handling out of our controller classes and into a model class. As we discussed on the previous page, the 
+controller class is not responsible for holding data.
```
gildedgardenia: I think this sentence comes off as confusing because the previous page makes a point of explaining that models don't hold data either.

src/chapters/aspdotnet-model-classes/model-class.rst
```diff
@@ -0,0 +1,224 @@
...
+
+In ``CodingEvents``, we'll remove the list data from ``EventsController`` and create a proper 
+C# class to deal with event items. We'll then update our action methods to take
+advantage of the new model and its properties, rather than the strings stored in the list.
+Lastly, because the controller is updating, the template variables it relies upon will also need to
```
gildedgardenia: I think this is supposed to be "updated"

src/chapters/aspdotnet-model-classes/model-class.rst
```diff
@@ -0,0 +1,224 @@
...
+            return Name;
+         }
+      }
+   }
+
+
+Now that we're working to move the data handling out from the controller classes and into a class of its own, 
+we'll need to update the ``POST`` handler that creates new events. 
+
+First, remove replace the ``Events`` dictionary with a list of ``Event`` objects.
```
gildedgardenia: "remove replace"

gildedgardenia (commented):
I like what is going on here! The explanations on this page are written at a high reading level, but once it gets to outlining the steps taken in the video, it becomes very readable for a wider audience.

src/chapters/aspdotnet-model-classes/model-class.rst
```diff
@@ -0,0 +1,214 @@
...
+         }
+
+         public override string ToString()
+         {
+            return Name;
+         }
+      }
+   }
+
+Now that our data is object-oriented, it’s quick and easy to add a new property affiliated with an event.
```
gildedgardenia, line 155: I would rephrase this to "Now that we are using the ``Event`` model, it's ..."

src/chapters/aspdotnet-model-classes/model-class.rst
```diff
@@ -0,0 +1,214 @@
...
+         }
+      }
+   }
+
+Now that our data is object-oriented, it’s quick and easy to add a new property affiliated with an event. 
+If we decide to add properties, such as ``Date`` or ``Location``, we can follow the pattern established. 
+Think about when we stored events as key-value pairs. At that stage, more significant changes were necessary 
+to add fields.
+
+In the ``Views`` folder, the ``Events/Add.cshtml`` template still uses a ``desc`` field so we don't need to update
```
gildedgardenia, line 160: Could you change the name now to ``description`` instead of ``desc``? After what you said about model binding, I thought I had named the two things differently, but this section tells us to name the property ``Description`` and the field ``desc``.

src/chapters/aspdotnet-model-classes/model-class.rst
```diff
@@ -0,0 +1,214 @@
...
+
+   Say we do add a ``Date`` property to the ``Event`` class. Which line would we add to ``Events/Index.cshtml`` 
+   to also display that value in our table of events?
+
+   #. ``<li>@evt.Date</li>``
+   #. ``<td>evt.Date</td>``
+   #. ``<td>@event.Date</td>``
+   #. ``<td>@evt.date</td>``
+
+.. ans: c, ``<td>@event.Date</td>``
```
gildedgardenia: According to what I read on this page, the answer would be ``<td>@evt.Date</td>``. Did I misread something?

src/chapters/aspdotnet-model-classes/model-class.rst
```diff
@@ -0,0 +1,214 @@
...
+      return Redirect("/Events");
+   }
+
+
+Check Your Understanding
+-------------------------
+
+.. admonition:: Question
+
+   True/False: Model code is framework independent.
```
gildedgardenia, line 193: I would rephrase this. It took me a second to get where you were going with this. I think we talk about so much being wrapped up in the framework, that I did a bit of a double-take. Just food for thought!

gildedgardenia (approved):
This page looks really great! I left some questions for you, but you can merge when you are ready!

src/chapters/aspdotnet-model-classes/model-class.rst
CarlyLanglois: i like this reference to oop - it reinforces the notion that models are just objects and emphasizes the relevance of the oop chapters that come before this

src/chapters/aspdotnet-model-classes/model-class.rst
CarlyLanglois: yeah, I'll keep this as desc for now and change the form field name in the model binding section. that way we have an opportunity to show why the property and form fields need to match. the work to make model binding happen is very simple so this adds some more context

src/chapters/aspdotnet-model-classes/model-class.rst
CarlyLanglois: i think a double take is good!

src/chapters/aspdotnet-model-classes/model-class.rst
CarlyLanglois: yes good catch

PR #49, opened 2020-05-15
Lsn11 model data layer page

CarlyLanglois:
3rd models intro page


src/chapters/aspdotnet-model-classes/data-layer.rst
```diff
@@ -0,0 +1,284 @@
+.. _data-layer:
+
+Models and Data 
+===============
+
+In order to work with data, we need to add another element to our MVC application. Say for example,
+we want to do things like remove an event from our list. Well, if two events both have the same name, 
+how might we identify which of those items to delete? We can't yet. So we need to tweak how we store
+event data.
```
gildedgardenia: This a confusingly worded paragraph. Maybe rephrase, "We need to add another element to our MVC application to effectively work with data. For example, if we want to remove an event from the list, but two events have the same name, we would not be able to identify which event we want to remove."

src/chapters/aspdotnet-model-classes/data-layer.rst
```diff
@@ -0,0 +1,284 @@
...
+===============
+
+In order to work with data, we need to add another element to our MVC application. Say for example,
+we want to do things like remove an event from our list. Well, if two events both have the same name, 
+how might we identify which of those items to delete? We can't yet. So we need to tweak how we store
+event data. 
+
+In ``CodingEvents``, we add a unique identifier field to ``Events`` to better handle and track distinct 
+``Event`` instances. We'll also create another model class called ``EventData`` to encapsulate data storage 
+and prepare ourselves for decoupling data from controller classes.
```
gildedgardenia: The first sentence very much ties to the example provided above, but as a student, I am not sure that I would understand based on the example why I need to create another model class.

src/chapters/aspdotnet-model-classes/data-layer.rst
```diff
@@ -0,0 +1,284 @@
...
+   As always, code along to the videos on your own ``CodingEvents`` project.
+
+.. index:: ! uid, ! unique identifier
+
+Add a Unique Id - Text
+----------------------
+
+Identifying data by a user-defined string called ``name`` is not a sustainable or scalable method
+of handling data in most situations. Consider the address book example. How can
+we distinguish between two contact entries with the same name field? It is a frequent
```
gildedgardenia, line 36: This secondary example does not make sense since you already gave an example earlier in the page. Maybe reference that to tie the page together. "As we saw above with our events list, we cannot know for sure which event to remove if two events have the same name."

src/chapters/aspdotnet-model-classes/data-layer.rst
```diff
@@ -0,0 +1,284 @@
...
+practice to add a **unique identifier** field (sometimes called, or even labelled, **uid**) to a class 
+responsible for modelling data. This ensures that our address book can contain two separate entries for 
+our contacts who have the same name as one another.
```
gildedgardenia, line 39: See the above comment. I would keep this in the context of events and ``CodingEvents``.

src/chapters/aspdotnet-model-classes/data-layer.rst
```diff
@@ -0,0 +1,284 @@
...
+         }
+
+         public override string ToString()
+         {
+            return Name;
+         }
+      }
+   }
+
+.. index:: ! get-only auto-implemented property
```
gildedgardenia: They have seen this syntax before in the context of ``readonly`` fields in Chapter 5.

gildedgardenia (commented):
This is looking good! I think the explanations at the start of the page need some clarification. A lot of the video text seems to count on students have a really good hold on the background before they start coding and some of your phrasing left me a little confused going into the videos.

src/chapters/aspdotnet-model-classes/data-layer.rst
CarlyLanglois: i changed this a bit but reminder that the previous two pages pretty strongly emphasize why want to do this and the example above is simply meant to supplement the notion of good data handling

src/chapters/aspdotnet-model-classes/data-layer.rst
CarlyLanglois: I see that this could feel repetitive but I see it more as a supplement - what about it isnt making sense for you?

src/chapters/aspdotnet-model-classes/data-layer.rst
CarlyLanglois: see the comment above - i don't think it would be harmful to add other contexts to why adding ids are good. we dont want to get stuck in the world of coding events, but instead reinforce that coding events is one example of how to implement these concepts

src/chapters/aspdotnet-model-classes/data-layer.rst
CarlyLanglois: yes, i thought about that too. but Im also thinking the label and definition are good here since we're giving a more in-depth look at what is taking place with the backing field

gildedgardenia: I agree! I would just add a reference back to the original context and an index entry for ``readonly``.

src/chapters/aspdotnet-model-classes/data-layer.rst
gildedgardenia: When I read through it, it feels a little bit like the question brought up in the intro goes unresolved. Whenever an author leaves a question unresolved and dives into another example, I find myself confused. However, if you pose the same question in the video and then resolve it there and this text is just providing a second example, I think that is fine!

src/chapters/aspdotnet-model-classes/data-layer.rst
gildedgardenia: Gotcha! I can see how this would work with the video accompanying it now!

gildedgardenia (approved):
This looks great! You are good to go!

PR #51, opened 2020-05-19
Lesson 11 Exercises

gildedgardenia:


src/chapters/aspdotnet-model-classes/exercises.rst
```diff
@@ -0,0 +1,101 @@
...
+
+.. admonition:: Tip 
+
+   As you work through these steps, test your code along the way! 
+   With each change you apply to your code, ask yourself what you expect to see when the application
+   is run. You may not find that all of the steps result in observable changes, though.
+   Use Visual Studio’s debugging tools and read your error messages if you run into issues after applying any of
+   the changes.
+
+#. Create the two handler methods listed below in ``EventsController``. We’ll add code
```
CarlyLanglois: change this to action methods

src/chapters/aspdotnet-model-classes/exercises.rst
```diff
@@ -0,0 +1,101 @@
...
+         public IActionResult Edit(int eventId = 1) {
+            // controller code will go here
+         }
+
+   #. Create a method to process the form with this signature:
+
+      .. sourcecode:: csharp
+         :linenos:
+
+         public IActionResult EditEvent(int eventId, string name, string desc) {
```
CarlyLanglois: we havent discussed a naming pattern to establish for get/post handler at the same route so lets talk about this in the working meeting tomorrow? chris's method in the older c# material is to use the same name for both methods and take advantage of attributes / the written order to have one Edit method map to GET at /Edit and another Edit method map to POST at /Edit. I like this because its what he did before and its simple and descriptive

src/chapters/aspdotnet-model-classes/exercises.rst
```diff
@@ -0,0 +1,101 @@
...
+#. Create the two handler methods listed below in ``EventsController``. We’ll add code
+   to these in a moment, so just put the outline in place for
+   now.
+
+   #. Create a method to display an edit form with this signature:
+
+      .. sourcecode:: csharp
+         :linenos:
+
+         public IActionResult Edit(int eventId = 1) {
```
CarlyLanglois: im having trouble with the default eventId value here. this method should be used when accessing the edit page for a particular event so is there going to be a scenario where a default eventId value is needed?

src/chapters/aspdotnet-model-classes/exercises.rst
```diff
@@ -0,0 +1,101 @@
...
+         :linenos:
+
+         public IActionResult EditEvent(int eventId, string name, string desc) {
+            // controller code will go here
+         }
+
+#. Add the necessary annotations to these methods for them to both live
+   at the path ``/Events/Edit``.
+    
+   #. Judging by the names of the handlers, which should handle ``GET`` requests and which should
```
CarlyLanglois: this note works for the java context but not so much for C#, no?

src/chapters/aspdotnet-model-classes/exercises.rst
```diff
@@ -0,0 +1,101 @@
...
+            // controller code will go here
+         }
+
+#. Add the necessary annotations to these methods for them to both live
+   at the path ``/Events/Edit``.
+    
+   #. Judging by the names of the handlers, which should handle ``GET`` requests and which should 
+      handle ``POST`` requests?
+
+   #. You’ll need to configure the route for ``Edit()`` to include the path variable ``eventId``,
```
CarlyLanglois: the solution code doesnt use a path variable at this time - im only able to reach an edit view for the default eventId value, 1. the way i understand this feature, the path should only be reachable with a user-provided event id. that value comes from the user selecting which event they intend to edit. In order to select that, the "edit event" button should be tied to an individual event table listing, rather than a view of all events to edit (ie, the way the delete functionality works right now). 
Can you clarify how you want this app to work or if im missing a bit of info? If I'm not missing something, the java solution might be helpful for spotting some differences - or I can clarify later

CarlyLanglois (changes requested):

src/chapters/aspdotnet-model-classes/exercises.rst
gildedgardenia: Done!

src/chapters/aspdotnet-model-classes/exercises.rst
gildedgardenia: I went back and removed the part about naming and added annotations so this should be clearer now!

src/chapters/aspdotnet-model-classes/exercises.rst
gildedgardenia: Removed!

src/chapters/aspdotnet-model-classes/exercises.rst
gildedgardenia: Fixed! The appropriate tag helper passes in the id as a query parameter so clarified that and adding the link as a column.

src/chapters/aspdotnet-model-classes/exercises.rst
gildedgardenia: Removed!

src/chapters/aspdotnet-model-classes/exercises.rst
CarlyLanglois: re: our convo in working meeting about naming - whats a way we can make this action method more explicit in what it does and its relationship to the get method above? maybe something like postEditEvent? or submitEditEvent?

src/chapters/aspdotnet-model-classes/exercises.rst
```diff
@@ -0,0 +1,97 @@
...
+         [HttpPost]
+         public IActionResult EditEvent(int eventId, string name, string desc) {
+            // controller code will go here
+         }
+
+#. Add the necessary annotations to these methods for them to both live
+   at the path ``/Events/Edit``.
+
+   #. You’ll need to configure the route for ``Edit()`` to include the query parameter ``eventId``, 
+      so that paths like ``/Events/Edit?id=3`` will work.
```
CarlyLanglois: this bit needs to direct the students to use a path variable instead of a query parameter. so paths like /events/edit/3 are what is used to edit events. 
for line 41 above, this means that the get handler wont be mapped at /events/edit, but events/edit/{id}. and correspondingly, the item on line 44 is meaning to address using a path variable

src/chapters/aspdotnet-model-classes/exercises.rst
```diff
@@ -0,0 +1,97 @@
...
+      set them as the values of the form fields.
+   
+   #. Add ``action="/events/edit"`` to the ``form`` tag.
+
+#. Add another input to hold the id of the event being edited. This
+   should be hidden from the user:
+
+   .. sourcecode:: html
+
+      <input type="hidden" value="@ViewBag.eventToEdit.Id" name="eventId" />
```
CarlyLanglois, line 73: it also be a good idea to add a note that this viewbag property will be whatever name they decide to give it, like this is just a suggestion and they shouldnt c/p it if they have named the ViewBag.eventToEdit something different

src/chapters/aspdotnet-model-classes/exercises.rst
```diff
@@ -0,0 +1,97 @@
...
+
+#. To access event editing, the user will need an edit option in the list of event data.
+
+   #. In ``Index.cshtml``, add a link to edit the 
+      event as a column in the event table:
+
+      .. sourcecode:: html
+         :linenos:
+
+        <td><a asp-controller="Events" asp-action="Edit" asp-route-id="@evt.Id">Edit Event</a></td>
```
CarlyLanglois: getting a build error on this sourcecode atm.
and i think its a good idea to add this explicit code since we haven't used this before, right? i think assigning the route-id portion of a route is a new concept right? can you consider adding more context around this since its an integral component of wiring up the edit paths created. I'll also suggest that this is a good candidate to be mentioned when you're introducing the AddEndpoints method back in the routing intro - that is where I understand this functionality to come from

CarlyLanglois (approved):
not a huge thing but the solution still uses the [fromForm] attributes

CarlyLanglois:
and reminder to pull in master updates

src/chapters/aspdotnet-model-classes/exercises.rst
gildedgardenia: Noted!

src/chapters/aspdotnet-model-classes/exercises.rst
gildedgardenia: I called it ``SubmitEditEventForm()``

src/chapters/aspdotnet-model-classes/exercises.rst
gildedgardenia: Done!

src/chapters/aspdotnet-model-classes/exercises.rst
gildedgardenia: Fixed!

PR #52, opened 2020-05-20
Lesson 11 Studio

gildedgardenia:


src/chapters/aspdotnet-model-classes/studio.rst
```diff
@@ -0,0 +1,115 @@
...
+
+#. Within the ``UserController``, create an action method with this signature:
+
+   .. sourcecode:: csharp
+
+      public IActionResult AddUser(User newUser, string verify) {
+         // add form submission handling code here
+      }
+
+   This will use model binding to create a new user object, ``newUser`, and
```
CarlyLanglois: newUser is missing one closing backtick

src/chapters/aspdotnet-model-classes/studio.rst
```diff
@@ -0,0 +1,115 @@
...
+      You don’t need to store the ``User`` object anywhere for this studio.
+      We’re focusing on form handling and validation in this exercise. If you
+      want to keep track of users using the method we employed in the models
+      lesson video, check out the Bonus Missions below.
+
+#. Check that the ``verify`` parameter matches the
+   password within the ``newUser`` object. If it does, render the
+   ``Index.cshtml`` view template with a message that welcomes the user by 
+   username. If the passwords don’t match, render the form again.
+
```
CarlyLanglois: this doesnt happen in the solution right now - im guessing because you've also found that viewbag does not carry data in a redirect? 
i think if this the case, we may want to restructure this studio since asking for usersignup and not delivering some view that verifies that a particular user is signed in feels odd, dont you think? what are your thoughts?

src/chapters/aspdotnet-model-classes/studio.rst
```diff
@@ -0,0 +1,115 @@
...
+Refining Form Submission
+------------------------
+
+#. Once registered, we want the user to access the form selecting their spa services. 
+
+   a. In ``Index.cshtml``, add an ``a`` element to take the user back to the path, ``/spa``, where the ``Spa/Index.cshtml`` template will be rendered.
+
+#. If the form is re-rendered when a password is not verified, we should let the user know that their form
+   was not properly submitted. Use ``ViewBag`` to add an ``error`` property, letting the user know
```
CarlyLanglois: this might be better phrased "add an error property to ViewBag.."

src/chapters/aspdotnet-model-classes/studio.rst
```diff
@@ -0,0 +1,115 @@
...
+--------------
+
+#. Add an ``Id`` property to ``User``, along with accessor methods (with
+   appropriate access level). Create a ``UserData`` class within
+   ``Data`` that provides access to a list of users via
+   ``Add()``, ``GetAll()``, and ``GetById()``.
+
+   a. In the ``Index.cshtml`` view, display a list of
+      all users by username. Each username should have a link that takes
+      you to a detail page that lists the user’s username and email.
```
CarlyLanglois: flagging this item related to passing the user data along - is this possible givne what we've introduced in the book so far?

CarlyLanglois (changes requested):
this is in pretty good shape but i think its important to confirm for the user that a user object has been created which Im not positive is possible with the instructions we have here but i could be missing a piece. have you been able to display the username on form submission?
another item that looks like tis missing to me is the spa services data - its kind of tangential to the objective of the studio but the services are missing from the spa menu view. have you caught that?

src/chapters/aspdotnet-model-classes/studio.rst
gildedgardenia: Done!

src/chapters/aspdotnet-model-classes/studio.rst
gildedgardenia: Figured out how to do it using a query parameter so it should work now!

src/chapters/aspdotnet-model-classes/studio.rst
gildedgardenia: Done!

src/chapters/aspdotnet-model-classes/studio.rst
gildedgardenia: I removed this bonus mission! It is possible, but it is way too large for a bonus mission!

src/chapters/aspdotnet-model-classes/studio.rst
```diff
@@ -0,0 +1,110 @@
...
+   pass it into your action method. 
+
+   .. admonition :: Note
+   
+      You don’t need to store the ``User`` object anywhere for this studio.
+      We’re focusing on form handling and validation in this exercise. If you
+      want to keep track of users using the method we employed in the models
+      lesson video, check out the Bonus Missions below.
+
+#. Check that the ``verify`` parameter matches the
```
CarlyLanglois, line 76: The studio looks good until some roadblocks around this step - 

the instructions above tell us to have the add user form post to /user. Which I think ultimately it should design-wise. we want that user index view to be the "welcome to the spa" message. the updated solution now though has the form posting to /user/add. 
I'm also thinking you can completely avoid that redirect viewbag issue by not returning a redirect and instead returning the View("Index"). Since we do briefly touch upon returning a different view than the default using an argument, this is a good opportunity to take advantage of it, no? It works and requires less workaround-code and less code overall than returning a redirect. A solution that returns View("Index") is more straightforward imo but i'd like to hear your opinion. And if its not clear, if you skip the redirect here, then you can just assign the viewbag username or name property in this method.

CarlyLanglois: the other thing is that I'd recommend renaming this action method based off of what we talked about in last week's working meeting. something more descriptive like, "processAddUserForm" or w/e. and i'd also say highly recommend we give them the instructions that this method will need information to respond to the add form post action and will need an explicit route attribute as well. maybe we don't need to include exactly what these look like in the sourcecode directive but some direction is a good idea - otherwise as I walk through the instructions myself, im worried these bits may get lost

src/chapters/aspdotnet-model-classes/studio.rst
```diff
@@ -0,0 +1,110 @@
...
+      You can do so using string concatenation: ``"/user?name=" + newUser.Username``. This will pass in the username as a query parameter.
+
+   b. Passing in the value to the ``Index.cshtml`` view also requires you to modify ``Index()``. You will need to add a parameter and use ``ViewBag``.
+
+Refining Form Submission
+------------------------
+
+#. Once registered, we want the user to access the form selecting their spa services. 
+
+   a. In ``Index.cshtml``, add an ``a`` element to take the user back to the path, ``/spa``, where the ``Spa/Index.cshtml`` template will be rendered.
```
CarlyLanglois: Just to add a few more details here - i'd clarify that the first index view mentioned is user/index. and probably not a bad idea to nudge them to use the asp-controller and asp-action directives, no?

the error processing instructions after this look good, btw. Just a suggestion though, can the solution code include smaller conditional blocks that display the error messages rather than repeating the whole form?

src/chapters/aspdotnet-model-classes/studio.rst
```diff
@@ -0,0 +1,110 @@
...
+   a. In the form submission action method, add the ``username`` and ``email`` fields of the submitted user as 
+      ``ViewBag`` properties. 
+   
+   b. Back in the form, add a value attribute to these form fields and make them equal to the
+      ``ViewBag`` properties. 
+
+Bonus Missions
+--------------
+
+#. Add a ``Date`` property in ``User``, and initialize it to the time the
```
CarlyLanglois, line 106: I see your point about shortening the goals of the bonus mission items but this feels too short imo. Can we add something to tie this new property back into the features of the app? Like previously i think the directive was to display the user's time added in some view?

CarlyLanglois (changes requested):

src/chapters/aspdotnet-model-classes/studio.rst
gildedgardenia: I made the following changes to the solution:

1. I called the action method ``SubmitAddUserForm()`` and it is now posting to the route: ``/user`` .
2. That method is using ``View()`` as you suggested. I couldn't remember if we covered explicitly that you can pass an object as a parameter to ``View()``, so I added a note about it. I know for sure we cover that in Lesson 12.

src/chapters/aspdotnet-model-classes/studio.rst
gildedgardenia: Clarified! I also followed your suggestion with the solution and I like it a lot. Thanks!

src/chapters/aspdotnet-model-classes/studio.rst
gildedgardenia: I added a second bonus mission to add an account detail div to the ``Index`` page. If they passed in simply the username in the earlier steps above, they will have to go back and refactor to pass in the whole User object. Otherwise, if they read the whole studio, they may have had the foresight to pass in the whole object. Going to add a comment to the studio that if a student passes in just the name that the studio will still be good since they only really need the whole object for the bonus mission!

src/chapters/aspdotnet-model-classes/studio.rst
CarlyLanglois: we definitely don't cover that to your second point. What I mean is to return View("Index")

CarlyLanglois (approved):

CarlyLanglois (changes requested):
sorry - i just approved this meaning to approve the wiki notes. lets just talk about what the solution code should look like when you're back next week?

PR #53, opened 2020-05-26
Lsn11 model binding

CarlyLanglois:
final page of model intros chapter


src/chapters/aspdotnet-model-classes/model-binding.rst
```diff
@@ -0,0 +1,126 @@
...
+
+.. index:: ! model binding
+
+We now introduce a useful technique to auto-create model instances, 
+called **model binding**. Model binding takes place when a whole 
+model object is created by the ASP.NET framework on form submission. This saves us the effort, 
+and the code, needed to pass in each form field to a controller. 
+
+Model binding reduces the amount of code we need to 
+write to create an object and it helps with validation (which we’ll explore further in the next
```
gildedgardenia: The "it" here is unnecessary.

gildedgardenia (approved):
Approved! I think this page is really well done! I think that model binding can be a deceptively simple concept and a lot of what goes on behind the scenes is what actually makes it work. Your explanations surrounding what model binding is and why we use it were clear and concise. This really helped me as a reader get the concepts down before heading into coding. When I started reading the supporting text for the videos, I didn't feel like I was lost. I felt like a good foundational understanding had already been built and we were building on top of it with code.

PR #59, opened 2020-06-18
SQL Part 2

gildedgardenia:


chrisbay (approved):
Looks good. There are a few build errors for this branch, but they all seem to be references to labels used in the first SQL chapter, so that's fine.

PR #71, opened 2020-06-26
Chapter 2 Updates

scottkuhl:
Added an example of boxing and unboxing.

Also, I did not add this because it might be covered later, but there is an additional type of comment that is meant to document the code and provide IntelliSense support.  You get this when you use three slashes above a method, class or property.

/// <summary>
///     Convert Fahrenheit into Celsius.
/// </summary>
public static void Main(string[] args)


gildedgardenia (approved):
Thanks Scott! I am approving this PR and merging in these changes!

PR #76, opened 2020-07-02
Public Properties vs Public Fields

scottkuhl:
I get this question from students a lot: Why would I use an auto-implemented property instead of just making a field public?

This change is my first attempt at explaining it without going into too much detail about how the compiler works and things like reflection.


gildedgardenia (approved):
Thanks for your contribution! I am approving this PR and merging in these changes!

PR #88, opened 2020-07-23
Rest exercises

CarlyLanglois:
rest exercises content  👍


gildedgardenia (approved):
I think this looks good! Two things: some sort of solution should be in Gitlab for the ease of use by course staff and would it be possible to introduce them to the API repo earlier in the chapter? These exercises felt like a lot to digest.

PR #96, opened 2020-09-03
SQL Studio Java to C# Changes

scottkuhl:
Replaced Java with C# references and updated Java style naming of properties to C# naming conventions.

Just a note:  I know the naming conventions used in this chapter are standard for MySQL.  But it will make it "slightly" harder to map the properties in C# since the names won't match.  But maybe that is the intention to teaching mapping to different standards in an ORM.


CarlyLanglois:
Thanks Scott!

PR #159, opened 2022-11-03
Chapter 20 Updates

gildedgardenia:


codinglikeagirl42 (approved):
Looks good.  I do have one suggestion - At the end of exercises page and at the beginning of the studio page there is a link to the installation Postman page.  It would be nice if there was a link back to the studio page.

PR #161, opened 2022-11-08
adding links for techjobs assignment demos

jwoolbright23:


gildedgardenia (approved):
These are the correct links. Before merging though, the search and list functions are not working on TechJobs MVC. When I click on those links, it just says unable to load this page. Can you look at this before merging?

